PR #31, opened 2022-05-15
refactor: error message improvement

fzn0x:
Use words and symbols that can be read well

Also lowercased `id` because the matched one was lowercase `id`.


rpaul-stripe (approved):
Seems reasonable, thanks!

PR #32, opened 2022-05-15
Add link title attribute to parser

### LINKED ISSUE #28: Parsing of link node doesn't return title attribute
### What happened?

I was trying to create a custom transform on "link" nodes to support adding additional custom attributes and in doing so, I noticed that the title attribute declared in markdown was never returned when calling Markdoc.parse.

In the Commonmark spec, [link-titles](https://spec.commonmark.org/0.30/#link-title) can be declared with the following syntax:

`[link](/uri "title")`

When running the parse function on a link with this syntax, however, only the href attribute is returned. No title attribute is ever returned from the parser.

Based on [the documentation for link node](https://markdoc.io/docs/nodes), I expected href and optionally the title attribute to be available in the node attributes when the markdown is parsed.

### To reproduce

The issue can be reproduced by running the following file.

```ts
import Markdoc, { Node } from "@markdoc/markdoc";

function printNodeAttributes(ast: Node) {
  if (ast.type === "link") {
    console.log(`Node type: ${ast.type}`);
    console.log(`Node attributes: ${JSON.stringify(ast.attributes)}`);
  }
  if (ast.children) {
    ast.children.forEach((child) => printNodeAttributes(child));
  }
}

const testLink = `[Markdoc](https://markdoc.io "markdoc link title")`;
const ast = Markdoc.parse(testLink);
printNodeAttributes(ast);
```

The output is 

```
Node type: link
Node attributes: {"href":"https://markdoc.io"}
```

I was expecting

```
Node type: link
Node attributes: {"href":"https://markdoc.io", title: "markdoc link title"}
```

### Version

0.1.1

### Additional context

It looks like the source of the issue may be [here](https://github.com/markdoc/markdoc/blob/d6e0ec5d1a4286212d0145eb301049386959855d/src/parser.ts#L32)

It looks like the handleAttrs function is only returning href for link nodes.

fzn0x:
Close #28

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -29,7 +29,7 @@ function handleAttrs(token: Token, type: string) {
       return { ordered: token.type.startsWith('ordered') };
     case 'link': {
       const attrs = Object.fromEntries(token.attrs);
-      return attrs?.title ? { href: attrs.href, title: attrs.title } : { href: attrs.href };
+      return { href: attrs.href, title: attrs.title };
     }
     case 'image': {
       const attrs = Object.fromEntries(token.attrs);
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -29,7 +29,7 @@ function handleAttrs(token: Token, type: string) {
       return { ordered: token.type.startsWith('ordered') };
     case 'link': {
       const attrs = Object.fromEntries(token.attrs);
-      return { href: attrs.href };
+      return attrs?.title ? { href: attrs.href, title: attrs.title } : { href: attrs.href };
     }
     case 'image': {
       const attrs = Object.fromEntries(token.attrs);


src/parser.ts
```diff
@@ -29,7 +29,7 @@ function handleAttrs(token: Token, type: string) {
       return { ordered: token.type.startsWith('ordered') };
     case 'link': {
       const attrs = Object.fromEntries(token.attrs);
-      return { href: attrs.href };
+      return attrs?.title ? { href: attrs.href, title: attrs.title } : { href: attrs.href };
```
RomanHotsiy, line 32: `Object.fromEntries` always returns an object so you don't need `?.`.

But also if you check the code around, no need to have a ternary at all, `undefined` value is just fine.

```suggestion
      return { href: attrs.href, title: attrs.title };
```

fzn0x: You are right since the `{ href: attrs.href }` has undefined `href` too.

rpaul-stripe (approved):
Thanks for contributing!

PR #40, opened 2022-05-18
Add footnote support

nvanexan:
Hi maintainers 👋

I had need for support for footnotes in my own project. Submitting this as a draft to add support for parsing footnotes via the [markdown-it-footnote](https://github.com/markdown-it/markdown-it-footnote) plugin.

I added a test to the parsers but feel free to play with it a bit. If you think this could be useful for all, feel free to comment, edit or add as you see fit. Thx!

Example text

```
 foo[^1]

[^1]: bar
```

Example output

```html
    <article>
      <p>
        foo<sup class="footnote-ref"><a id="fnref1" href="#fn1">1</a></sup>
      </p>
      <section class="footnotes">
        <ol class="footnotes-list">
          <li class="footnote-item" id="fn1">
            <p>bar<a class="footnote-anchor" href="#fnref1">↩</a></p>
          </li>
        </ol>
      </section>
    </article>
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,13 +1,12 @@
 {
   "name": "@markdoc/markdoc",
-  "version": "0.0.1",
+  "version": "0.1.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@markdoc/markdoc",
-      "version": "0.0.1",
-      "hasInstallScript": true,
+      "version": "0.1.2",
       "license": "MIT",
       "devDependencies": {
         "@types/jasmine": "^3.10.2",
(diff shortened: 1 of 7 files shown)


nvanexan:
Updated to remove the commits in a previous PR and to get the build check to pass. Cheers! 🍻

rpaul-stripe:
I appreciate that you took the time to put this PR together, but I think we're going to pass on this one. We'd prefer not to introduce additional syntax or dependencies on additional markdown-it plugins, especially in cases where it would be practical to implement the feature with Markdoc tags.

matildepark:
> especially in cases where it would be practical to implement the feature with Markdoc tags

So are we to advise writers to use `{% footnote %}` format instead of just `[^1]`?

matildepark:
For future readers, I ended up porting the work in nvanexan/nve2022#7 for my situation (double kudos to OP here).

PR #45, opened 2022-05-19
Add support for image titles

nvanexan:
Adding support for image titles per CommonMark Spec: https://spec.commonmark.org/0.30/#example-571

This can be used downstream to wrap images with `<figure>` elements and with `<figcaption>` elements for the title where the attributes return a title.

Example:

```md
![foo](/url "title")
```

```html
<figure>
   <img src="/url" alt="foo" />
   <figcaption>title</figcaption>
</figure>
```

### DIFF
--- a/src/parser.test.ts
+++ b/src/parser.test.ts
@@ -146,6 +146,22 @@ describe('Markdown parser', function () {
         content: 'This is a test\n',
       });
     });
+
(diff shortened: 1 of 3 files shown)


rpaul-stripe (approved):
Thanks for contributing! This looks good to me. 👍

PR #49, opened 2022-05-24
Add support for table alignment

thcyron:
In GFM, alignment in tables can be specified using colons in the delimiter line:

```
| Left | Center | Right |
| :--- | :----: | ----: |
| Left | Center | Right |
```

This PR updates to parser to carry over the `text-align` style property to the node’s `align` attribute.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -48,6 +48,19 @@ function handleAttrs(token: Token, type: string) {
         ? { content: token.content }
         : { content: token.content, language };
     }
+    case 'td':
+    case 'th': {
+      const attrs = Object.fromEntries(token.attrs);
+      let align = 'left';
+      if (attrs.style) {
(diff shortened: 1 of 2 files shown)


rpaul-stripe (approved):
This is great, thanks for contributing!

src/parser.ts
```diff
@@ -48,6 +48,19 @@ function handleAttrs(token: Token, type: string) {
         ? { content: token.content }
         : { content: token.content, language };
     }
+    case 'td':
+    case 'th': {
+      const attrs = Object.fromEntries(token.attrs);
+      let align = 'left';
```
mfix-stripe, line 54: @rpaul-stripe we probably want the make this `undefined` and then set it to `left` with `:--` right? Just so `left` doesn't always show up in the tree. Or do you prefer it this way?

Issue #56, opened 2022-05-28
(question) Can customized React components are used instead of default components?


matsudai:
Currently, components that start with a lowercase letter are not customizable.

Is there a good way to customize "h2", for example?

```jsx
import { parse, renderers, transform } from "@markdoc/markdoc";

const node = parse('## Header 2');

const readableTreeNode = transform(node);

const component = renderers.react(readableTreeNode, React, {
  components: {
    // Ignored.
    h2: (props) => <h2 {...props} style={{ color: "red" }} />,
  },
});
```

Thanks.

---

I have tried these solutions.

**(1) Map components to components that start with UPPERCASE letter.**

```jsx
import { nodes, parse, renderers, Tag, transform } from "@markdoc/markdoc";

const node = parse(`## Header 2`);

// Thanks for https://markdoc.io/docs/nodes#customizing-markdoc-nodes
const readableTreeNode = transform(node, {
  nodes: {
    heading: {
      ...nodes.heading,
      // Required to transform from "heading" to "Heading".
      transform: (node, config) => {
        const attributes = node.transformAttributes(config);
        const children = node.transformChildren(config);
        return new Tag(`Heading${node.attributes.level}`, attributes, children);
      },
    },
  },
});

const component = renderers.react(readableTreeNode, React, {
  components: {
    // OK. (However, required to transform from heading/level2 to H2)
    Heading2: (props: any) => <h2 {...props} style={{ color: "blue" }} />,
  },
});
```

**(2) Change the priority of using default components, customized components and the function.**

(Update to use the default component only if no customized component is found.)

Swap `return name` and `return components[name]` in `src/renderers/react/shared.ts`

PR #57, opened 2022-05-28
Update to use customized React component instead of default component. #56

matsudai:
This change is the solution (2) of https://github.com/markdoc/markdoc/issues/56

> (2) Change the priority of using default components, customized components and the function.
>
> (Update to use the default component only if no customized component is found.)
>
> Swap `return name` and `return components[name]` in `src/renderers/react/shared.ts`

### DIFF
--- a/src/renderers/react/shared.ts
+++ b/src/renderers/react/shared.ts
@@ -8,9 +8,7 @@ export function tagName(
 ): string | Component {
   return typeof name !== 'string'
     ? 'Fragment'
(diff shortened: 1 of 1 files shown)


matsudai:
Sorry. This solution is bad.

In the following cases, the behavior is different from the current one.

1. `name` starts with lowercase.
2. `components` is a function.
3. `components` function does not return a component responding to 1.

matsudai:
@mfix-stripe Thanks for your answer #64 .

PR #83, opened 2022-06-09
Test cases for CommonMark code_block node

### LINKED ISSUE #82: Enhancement / Bug: Missing CommonMark node type 'code_block'
### What happened?

I added a (test) markdown file that had the following expressions:

```markdown

`inline code!`

~~~text
tilde-fenced-code!
~~~

    # https://spec.commonmark.org/0.30/#indented-code-block
    4-space indented code

```

When I rendered it with the react renderer under nextjs, this document threw the error:

`Error: Undefined node 'code_block'`

### To reproduce

1. Enter the content in a markdown file as described above
1. Render it in your markdoc-based nextjs site
1. See the error!

Will submit a PR with test cases shortly.

### Version

_No response_

### Additional context

I expected this to be supported because the documentation says that the nodes supported are the CommonMark nodes.

That said, it is indeed true that the exact nodes supported table does not include 'code_block' so the documentation may be "technically correct"

Workarounds include:
- Provide my own code_block implementation in my schema
- Avoid validating my markdown document and just swallow the error

fbartho:
Closes #82

`markdown-it` related rule: https://github.com/markdown-it/markdown-it/blob/e5986bb7cca20ac95dc81e4741c08949bf01bb77/lib/rules_block/code.js

### DIFF
--- a/spec/marktest/tests.yaml
+++ b/spec/marktest/tests.yaml
@@ -148,6 +148,19 @@
       attributes: { id: 'test', 'data-language': 'ruby' }
       children: ["example\n"]
 
+# https://spec.commonmark.org/0.30/#indented-code-block
+- name: code_block node via indentation
+  code: "\n\n    Code!\n\n"
+  expected:
+    - tag: pre
+      children: ['Code!']
+
+- name: code_block node via tab indentation
+  code: "\n\n\tCode!\n\n"
+  expected:
+    - tag: pre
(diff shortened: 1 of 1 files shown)


mfix-stripe:
Hey @fbartho, thanks for submitting this—really appreciate it. @rpaul-stripe is OOO at the moment, but I will chat with the rest of the team and get back to this thread ASAP.

PR #91, opened 2022-06-15
Refactor transformer types

mfix-stripe:
PTAL @rpaul-stripe 

* Expose `globalAttributes`
* Move `Transformer` type to `types.ts`

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -8,7 +8,7 @@ import tags from './src/tags/index';
 import { truthy } from './src/tags/conditional';
 import functions from './src/functions/index';
 import renderers from './src/renderers/index';
-import transformer from './src/transformer';
+import transformer, { globalAttributes as attributes } from './src/transformer';
 import validator from './src/validator';
 import { parseTags } from './src/utils';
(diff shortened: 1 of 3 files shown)


index.ts
```diff
@@ -129,6 +130,7 @@ export {
   nodes,
   tags,
   functions,
+  attributes,
```
matv-stripe, line 133: what is the use case for exposing  this?

```
export const globalAttributes: AttributesSchema = {
  class: { type: Class, render: true },
  id: { type: String, render: true },
};
```

matv-stripe (commented):
quick question, @mfix-stripe

index.ts
mfix-stripe: Just to align it with the other primitive collections: `nodes`, `tags`, `functions`. Do you think we should call it `globalAttributes` instead?

mfix-stripe:
PTAL @rpaul-stripe — WDYT about exporting this config? Should we call it `globalAttributes` or `attributes`?

rpaul-stripe (approved):

PR #93, opened 2022-06-16
Patch parseTags

### LINKED ISSUE #72: Strings starting with "{%" inside code blocks are parsed as unclosed Markdoc, clogging builds
### What happened?

I've been migrating our website to Markdoc. We use a bit of an esoteric programming language that does use `{%term}` format at times. When running `next build`, on the build would halt on one of 5 pages and my node would gradually use 12GB of RAM until my system froze.

Deleting files and escaping the Markdown render I've nailed it down to pages that have strings that start with `{%` inside a code block. Examples:

image

image

image

If I delete these lines from all these files, the build runs fine. If I retain even one, it goes into its death spiral.

### To reproduce

Place one of those strings on a page. Should be enough. It looks like the Markdoc website wraps examples in a custom tag — was this an intentional workaround? We can do similar ...

### Version

0.1.2

### Additional context

_No response_

mfix-stripe:
PTAL @matv-stripe 

Closes #72

### DIFF
--- a/src/utils.test.ts
+++ b/src/utils.test.ts
@@ -165,8 +165,21 @@ describe('Templating', function () {
               example=1 another="test}"}
         `;
 
-        const end = findTagEnd(example, 0, '{', '}');
+        const end = findTagEnd(example, 0);
         expect(end).toEqual(undefined);
+      });
+      it("shouldn't hang when {% is included in code block", () => {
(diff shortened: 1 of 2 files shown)


mfix-stripe:
Closing in favor of #94

PR #94, opened 2022-06-16
Patch parseTags to avoid crashing on unopened tag

matv-stripe:
## Problem

In `parseTags` which is called in fence blocks we parse out Markdoc tags: https://github.com/markdoc/markdoc/blob/main/src/tokenizer/plugins/annotations.ts#L127

This function has a bug where it goes in an infinite loop when there is an opening, but not a closing tag.

Closes https://github.com/markdoc/markdoc/issues/72
Closes #93 

## Solution
Patch


src/utils.ts
```diff
@@ -70,7 +72,14 @@ export function parseTags(content: string, firstLine = 0): Token[] {
 
     if (!content.startsWith(OPEN, pos)) continue;
 
-    const end = findTagEnd(content, pos) || 0;
```
matv-stripe, line 73: it felt odd to me to treat a non-found tag as position `0`. I think that breaks assumptions, so I thought an early continue would make more sense

matv-stripe (commented):
r? @mfix-stripe

mfix-stripe (approved):

PR #100, opened 2022-06-20
Fix TLD in README Shoutout

MarcusSorealheis:
Currently the readme points to https://markdoc.dev, which points to a non-existent domain. It could be that this is the intended domain, yet the public-facing site is hosted at the .io TLD. If that is the case, the .dev could reroute to the .io. Otherwise, you could decide to call out that you decided to go to .io after the fact. There are many options. :)

I'm open to whatever you decide, but thought some fix should land.


mfix-stripe (approved):

mfix-stripe:
Thanks @MarcusSorealheis!

PR #104, opened 2022-06-24
[RFC] Make default export a Markdoc class component

mfix-stripe:
@rpaul-stripe WDYT about this? 

Cc @markdoc/core


rpaul-stripe (approved):
Much cleaner. 👍

PR #105, opened 2022-06-24
Disable indented code_block

### LINKED ISSUE #82: Enhancement / Bug: Missing CommonMark node type 'code_block'

mfix-stripe:
Closes #82
Closes #83

### DIFF
--- a/src/tokenizer/index.ts
+++ b/src/tokenizer/index.ts
@@ -12,7 +12,11 @@ export default class Tokenizer {
     this.parser = new MarkdownIt(config);
     this.parser.use(annotations, 'annotations', {});
     this.parser.use(frontmatter, 'frontmatter', {});
-    this.parser.disable('lheading');
+    this.parser.disable([
+      'lheading',
+      // Disable indented `code_block` support https://spec.commonmark.org/0.30/#indented-code-block
+      'code',
+    ]);
   }
 
   tokenize(content: string): Token[] {
--- a/src/validator.test.ts
+++ b/src/validator.test.ts
@@ -295,4 +295,12 @@ describe('validate', function () {
       expect(output).toEqual([]);
     });
   });
+
+  it('should not error for missing support for code_block', () => {
+    const example = `   # https://spec.commonmark.org/0.30/#indented-code-block
+    4-space indented code`;
+    const output = validate(example, {});
+
+    expect(output).toEqual([]);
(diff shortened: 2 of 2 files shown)


src/tokenizer/index.ts
```diff
@@ -12,7 +12,11 @@ export default class Tokenizer {
     this.parser = new MarkdownIt(config);
     this.parser.use(annotations, 'annotations', {});
     this.parser.use(frontmatter, 'frontmatter', {});
-    this.parser.disable('lheading');
+    this.parser.disable([
+      'lheading',
+      // Disable indented `code_block` support https://spec.commonmark.org/0.30/#indented-code-block
+      'code',
```
matv-stripe, line 18: does this disable regular fence code blocks too?

image

matv-stripe (commented):
question

src/tokenizer/index.ts
mfix-stripe: Fences are supported by this rule: https://github.com/markdown-it/markdown-it/blob/master/lib/rules_block/fence.js

`code_block` nodes are supported by this rule (with name: `code`): https://github.com/markdown-it/markdown-it/blob/master/lib/rules_block/code.js

matv-stripe: huh. That is confusing 😆

matv-stripe (approved):

fbartho:
Oh! So indented code_block was intentionally disabled? I didn’t catch that in docs!

PR #117, opened 2022-07-05
Store annotations as [] in AST

mfix-stripe:
WDTY @rpaul-stripe 

Cc @matv-stripe @nkohari-stripe

### DIFF
--- a/src/ast/node.test.ts
+++ b/src/ast/node.test.ts
@@ -4,6 +4,7 @@ import Node from './node';
 import Variable from './variable';
 
 import Tag from '../ast/tag';
+import { json } from 'stream/consumers';
 
 describe('Node object', function () {
   describe('traversal', function () {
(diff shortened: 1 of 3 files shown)


src/ast/node.test.ts
```diff
@@ -305,6 +305,35 @@ describe('processor rendering', function () {
     });
   });
 
+  describe('annotations', () => {
+    it('multiple values should be ordered correctly', () => {
+      const example = markdoc.parse(
+        `\`\`\`js {% z=true .class y=2 x="1" #id %} \nContent\n\`\`\``
```
matv-stripe, line 311:
```
## Heading {% anchored=true %}
```

does `{% anchored=true %}` also go through the same annotations section?

matv-stripe (commented):
I think this makes sense. I don't think it adds too much weight to the AST. 


It does feel a little weird to me to have the duplication between `annotations` and `attributes` though

PR #124, opened 2022-07-15
Add Partial file attribute validation

### LINKED ISSUE #122: No validation on partials
### What happened?

Using a partial that doesnt exist fails silently, I don't know if this is intended, but I think since integrity of the docs depends on every piece "working" partials should have a validation error.

Is this intended for some specific reason?

I ended up adding the validation on my side, but if this is something that could go inside of the default functionality I'll be glad to make a PR.

```typescript
import { tags } from '@markdoc/markdoc';

const config: Config = {
    tags: {
        ...tags,
        partial: {
            ...tags.partial,
            validate: (node, config) => {
                const { partials = {} } = config;
                const { file } = node.attributes;
                const partial: Node | Node[] = partials[file];

                if (!partial) {
                    return [
                        {
                            id: 'partial-children',
                            level: 'critical',
                            message: `Partial ${file} not found`,
                            location: node.location
                        }
                    ];
                }

                return [];
            }
        }
    }
};

```

### To reproduce

```typescript
const doc = `{% partial file="non-existent.md" /%}`;
const ast = Markdoc.parse(doc);
const errors = Markdoc.validate(ast, config);
console.log(errors) // []
```

### Version

0.1.3

### Additional context

_No response_

Thank you! I think this library is very interesting, I'm currently evaluating if it fits our needs

mfix-stripe:
Closes #122

PTAL @rpaul-stripe 
Cc @ntgussoni @matv-stripe

### DIFF
--- a/src/tags/partial.ts
+++ b/src/tags/partial.ts
@@ -1,12 +1,30 @@
-import type { Node, Config, Schema } from '../types';
+import type { Node, Config, Schema, ValidationError } from '../types';
+
+class File {
+  validate(file: any, config: Config): ValidationError[] {
+    const { partials = {} } = config;
+    const partial = partials[file];
+
+    if (!partial)
+      return [
+        {
+          id: 'attribute-value-invalid',
+          level: 'error',
+          message:
+            "The 'file' attribute value must be set in `config.partials`",
+        },
+      ];
+
+    return [];
+  }
+}
 
 export const partial: Schema = {
   selfClosing: true,
   attributes: {
-    file: { type: String, render: false, required: true },
+    file: { type: File, render: false, required: true },
     variables: { type: Object, render: false },
   },
-
   transform(node: Node, config: Config) {
     const { partials = {} } = config;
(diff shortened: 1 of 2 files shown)


src/tags/partial.ts
```diff
@@ -1,12 +1,30 @@
...
+    const { partials = {} } = config;
+    const partial = partials[file];
+
+    if (!partial)
+      return [
+        {
+          id: 'attribute-value-invalid',
+          level: 'error',
+          message:
+            "The 'file' attribute value must be set in `config.partials`",
```
mfix-stripe, line 14: Feedback on verbiage welcome!

ntgussoni:
That was fast! Thanks a lot!

src/tags/partial.ts
ntgussoni: Ah, nvm.. I just saw you changed it

mfix-stripe:
> That was fast! Thanks a lot!

Thanks for the awesome bug report!

PR #134, opened 2022-07-19
Add validation for functions existing

### LINKED ISSUE #123: No error when accessing missing variables
I see the same happens when accessing missing variables, isn't this expected to throw an error?

_Originally posted by @ntgussoni in https://github.com/markdoc/markdoc/issues/122#issuecomment-1185614546_

mfix-stripe:
Closes #123

PTAL @rpaul-stripe

### DIFF
--- a/src/ast/base.ts
+++ b/src/ast/base.ts
@@ -3,6 +3,7 @@ import type {
   Config,
   Function as AstFunction,
   Scalar,
+  Variable as AstVariable,
 } from '../types';
 
 export function isAst(value?: any): value is AstType {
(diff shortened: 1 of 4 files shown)


rpaul-stripe:
In the Stripe documentation, we compute the variables that we pass into the Markdoc environment per request because many of the values are specific to the individual user. That means we don't have the variables available in any of the places where we perform validation internally (CI or our VSCode extension). We would expect `config.variables` to be undefined in those environments and don't want that to be treated as an error.  

Maybe we could make it so that variable validation only occurs if `config.variables` exists?

mfix-stripe:
> In the Stripe documentation, we compute the variables that we pass into the Markdoc environment per request because many of the values are specific to the individual user. That means we don't have the variables available in any of the places where we perform validation internally (CI or our VSCode extension). We would expect `config.variables` to be undefined in those environments and don't want that to be treated as an error.
> 
> Maybe we could make it so that variable validation only occurs if `config.variables` exists?

@rpaul-stripe this validation is only checking that the variables object has the property, not if the value is correct, so even if we keep this PR as is I believe it will still work. 

Though I am totally peaceful with only validating if `config.variables`.

mfix-stripe:
@rpaul-stripe just updated this to only validate if `config.variables` exists—PTAL

segphault (approved):
Thanks!

rpaul-stripe (approved):

PR #139, opened 2022-07-23
Pass node to validations so that location can be accessed for errors

ntgussoni:
This PR makes sure you get a reference to the node in validations so that you can return errors with more information eg: location.

I see there are some typing that could be improved eg: `export class Class implements CustomAttributeTypeInterface {` idk why its not typing `validate` correctly.  But I didn't want to change too much.

I made a breaking change, if you will, `validate(file: any, node: Node, config: Config)` although it didn't break any tests. I think it would be good practice to pass a single property as an object to avoid having to do any refactoring or breaking change.

Please let me know if you want me to change anything (or if this PR doesn't make sense).

### DIFF
--- a/src/schema-types/class.ts
+++ b/src/schema-types/class.ts
@@ -1,7 +1,7 @@
-import { CustomAttributeTypeInterface, ValidationError } from '../types';
+import { CustomAttributeTypeInterface, ValidationError, Node } from '../types';
 
 export class Class implements CustomAttributeTypeInterface {
-  validate(value: any): ValidationError[] {
+  validate(value: any, node: Node): ValidationError[] {
     if (typeof value === 'string' || typeof value === 'object') return [];
 
     return [
@@ -9,6 +9,7 @@ export class Class implements CustomAttributeTypeInterface {
         id: 'attribute-type-invalid',
         level: 'error',
         message: `Attribute 'class' must be type 'string | object'`,
+        location: node.location,
       },
     ];
   }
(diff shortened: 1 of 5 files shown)


mfix-stripe:
@ntgussoni, I think for now, I would prefer that we just automatically pass `node.location` from the node that the attribute failure happens on, instead of passing it in an requiring that users set it themselves in every validation function. This would also avoid a potential breaking change. 

WDYT @rpaul-stripe?

ntgussoni:
Yeah after checking a little bit more I realized it was being set, although it loses a bit of encapsulation if you have do another run on the after calling validate in order to get better descriptions and point to the right line

mfix-stripe:
> Yeah after checking a little bit more I realized it was being set, although it loses a bit of encapsulation if you have do another run on the after calling validate in order to get better descriptions and point to the right line.

@ntgussoni yes, we absolutely want to be able to point to the right line, but I'm not sure we should handle this in userland. I'm not uniformly opposed to passing `node` to custom attribute `validate` either, I just want to have a really good reason to. And if we do, we wouldn't make it a breaking change, so it would have to be passed as the 3rd parameter.

rpaul-stripe:
The top-level Markdoc `validate` function [returns the location data](https://github.com/markdoc/markdoc/blob/main/index.ts#L91-L95) alongside the error. I don't think we need to add the location to the error itself. Is there a specific case where the current behavior is not sufficient? I think we can probably close this.

ntgussoni:
Fair enough, i'll close this. Thanks a lot

PR #142, opened 2022-07-23
makes parser public so that it can be extended

ntgussoni:
In order to add custom plugins we need to call `.use` coming out of markdownIt.

This PR simply makes parser public so that it's accessible.

### DIFF
--- a/src/tokenizer/index.ts
+++ b/src/tokenizer/index.ts
@@ -4,10 +4,12 @@ import frontmatter from './plugins/frontmatter';
 import type Token from 'markdown-it/lib/token';
 
 export default class Tokenizer {
-  private parser: MarkdownIt;
+  parser: MarkdownIt;
 
   constructor(
-    config: MarkdownIt.Options & { allowIndentation?: boolean } = {}
+    config: MarkdownIt.Options & {
+      allowIndentation?: boolean;
+    } = {}
   ) {
     this.parser = new MarkdownIt(config);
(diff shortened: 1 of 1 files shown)


ntgussoni:
Just to add to this, I tried to make my own custom tokenizer, just so I get unblocked, but the stock plugins are not being exported. That would be ok for me now as well

ntgussoni:
> Hey @ntgussoni, thanks for the submission. Unfortunately, we are currently treating markdown-it as an explicit internal dependency (as you can see by our package.json), and have thus chosen to keep those details private. If you wish to access the `parser`, it is possible to do so, you just need to mark the tokenizer as `any`. Given this, please do so at your own risk! 🙂

Thanks for the reply! I see you don't expect to add more syntax to the current, but not having an escape hatch for people that wish to add more functionality (at their own risk) might turn down some adopters.

Is there any roadmap or more detailed vision that we can see? I don't want to keep bothering you with this stuff if it's not something is in the the plan

mfix-stripe:
> but not having an escape hatch for people that wish to add more functionality (at their own risk) might turn down some adopters.
@ntgussoni there is an escape hatch, as described above. We just want the inherent risk to be encoded into that escape hatch, via the unsafe types. Hope that makes sense to you. 

As for the roadmap, that is a great question. Once we get firm roadmap plans together, we will share them as a GitHub project.

PR #174, opened 2022-08-17
Add test for auto-closing tags

mfix-stripe:


src/formatter.test.ts
```diff
@@ -143,6 +152,12 @@ describe('Formatter', () => {
     const expected = `
 {% key /%}
 
+{% a /%}
```
mfix-stripe, line 155: @rpaul-stripe is the desired behavior for tags without children?

mfix-stripe (commented):
PTAL @rpaul-stripe

rpaul-stripe (approved):
Yes, that seems like the right behavior to me.

PR #182, opened 2022-08-23
Improve error message for invalid attribute and add test

matv-stripe:
## Problem

The error message for attribute invalidity could be improved by adding the value that was passed in. This would make some debugging easier.

In my use case, I have a glossary term that is invalid:

```
 {
      error: {
        id: 'attribute-value-invalid',
        level: 'error',
        message: `Attribute 'term' must match one of ["3D-Secure","3D-Secure-2", ...","wire-transfer"]`
      },
  }
```

It would be nice if it said what term was passed in. It would make cmd+f to replace the error easier instead of having to jump to the line number and look for it.

## Solution

Update copy + add test

### DIFF
--- a/src/validator.test.ts
+++ b/src/validator.test.ts
@@ -234,7 +234,7 @@ describe('validate', function () {
 
   describe('attribute validation', () => {
     it('should return error on failure to match array', () => {
-      const example = '{% foo bar=baz() /%}';
+      const example = '{% foo jawn="cat" /%}';
       const schema = {
         tags: {
           foo: {
@@ -248,7 +248,7 @@ describe('validate', function () {
         },
       };
 
-      expect(validate('{% foo jawn="cat" /%}', schema)).toDeepEqualSubset([
+      expect(validate(example, schema)).toDeepEqualSubset([
         {
           type: 'tag',
           error: {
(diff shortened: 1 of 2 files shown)


src/validator.test.ts
```diff
@@ -232,6 +232,34 @@ describe('validate', function () {
     });
   });
 
+  describe('attribute validation', () => {
+    it('should return error on failure to match array', () => {
```
mfix-stripe, line 236: If you are up for it, it might make sense to move the other attribute validation tests into this `describe` block

mfix-stripe (approved):
This is great 🙌 ! One optional comment

src/validator.test.ts
```diff
@@ -232,6 +232,34 @@ describe('validate', function () {
...
+          },
+        },
+      };
+
+      expect(validate(example, schema)).toDeepEqualSubset([
+        {
+          type: 'tag',
+          error: {
+            id: 'attribute-value-invalid',
+            message: `Attribute 'jawn' must match one of ["bar","baz","bat"]. Got 'cat' instead.`,
```
mfix-stripe, line 256: I can't even with this "jawn" variable 🤣

src/validator.test.ts
matv-stripe: makes sense, I will do in another PR since it involves factoring out the schema file

PR #198, opened 2022-09-15
Adds comment syntax to Markdoc

rpaul-stripe:
This PR introduces support for comments in Markdoc. Some users (including Stripe) have previously relied on a no-op `{% comment %}` tag to be able to include non-rendering content in a document, but the content inside of the tag is still subject to parsing and validation, which creates some undesirable side effects, particularly when trying to comment out portions of a doc that contain other tags or malformed tags.

Rather than special-casing the behavior of the `{% comment %}` tag in a way that would undermine the consistency of tag semantics, we instead decided that Markdoc comments will use the standard HTML comment syntax (`<!-- -->`). This syntax is already supported for commenting in many CommonMark-compliant parsers, via broader support for arbitrary HTML markup inside of Markdown.

This PR adds a new plugin to MarkdownIt that just adds that comment syntax, making it possible to have HTML comments without having to enable general HTML parsing. The resulting comments are incorporated into the AST but are omitted from the rendered document (they do not get propagated through as HTML comments in the output). Users can override the schema definition for comment nodes and implement their own transform function if they would like different behavior, though we do not have support for rendering actual HTML comments via the render tree.

Things included in this PR:

* A MarkdownIt plugin that introduces block and inline parser rules for Markdoc comments 
* A new document node schema type for comments
* Schema updates to allow comment nodes at the document level and in inlines (I think this is all we need, but I may have missed something here)
* Test cases for the parser and for the MarkdownIt plugin
* Marktest cases to ensure that comments parse and do not render as expected
* A tweak to the parser code to make sure that the content of the comment gets applied to the AST node
* A new tokenizer option for turning on the plugin
* An assortment of unrelated marktest cases that I wrote to test various things recently and rolled into this PR

### DIFF
--- a/spec/marktest/index.ts
+++ b/spec/marktest/index.ts
@@ -18,7 +18,10 @@ class Loader extends yaml.loader.Loader {
   }
 }
 
-const tokenizer = new markdoc.Tokenizer({ allowIndentation: true, allowComments: true });
+const tokenizer = new markdoc.Tokenizer({
+  allowIndentation: true,
+  allowComments: true,
+});
 
 function parse(content: string, file?: string) {
   const tokens = tokenizer.tokenize(content);
--- a/spec/marktest/index.ts
+++ b/spec/marktest/index.ts
@@ -18,7 +18,7 @@ class Loader extends yaml.loader.Loader {
   }
 }
 
-const tokenizer = new markdoc.Tokenizer({ allowIndentation: true });
+const tokenizer = new markdoc.Tokenizer({ allowIndentation: true, allowComments: true });
 
 function parse(content: string, file?: string) {
   const tokens = tokenizer.tokenize(content);
--- a/src/parser.test.ts
+++ b/src/parser.test.ts
@@ -7,7 +7,7 @@ import { any } from 'deep-assert';
 
 describe('Markdown parser', function () {
   const fence = '```';
-  const tokenizer = new Tokenizer({allowComments: true});
+  const tokenizer = new Tokenizer({ allowComments: true });
 
   function convert(example) {
     const content = example.replace(/\n\s+/gm, '\n').trim();
@@ -637,7 +637,7 @@ describe('Markdown parser', function () {
     }).not.toThrow();
   });
 
-  it('parsing comments', function() {
+  it('parsing comments', function () {
     const example = convert(`
     this is a test
 
@@ -647,9 +647,9 @@ describe('Markdown parser', function () {
     expect(example).toDeepEqualSubset({
       type: 'document',
       children: [
-        {type: 'paragraph'},
-        {type: 'comment', attributes: {content: 'foo'}}
-      ]
+        { type: 'paragraph' },
+        { type: 'comment', attributes: { content: 'foo' } },
+      ],
     });
   });
 });
--- a/src/parser.test.ts
+++ b/src/parser.test.ts
@@ -7,7 +7,7 @@ import { any } from 'deep-assert';
 
 describe('Markdown parser', function () {
   const fence = '```';
-  const tokenizer = new Tokenizer();
+  const tokenizer = new Tokenizer({allowComments: true});
 
   function convert(example) {
     const content = example.replace(/\n\s+/gm, '\n').trim();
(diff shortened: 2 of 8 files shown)


spec/marktest/index.ts
```diff
@@ -18,7 +18,10 @@ class Loader extends yaml.loader.Loader {
   }
 }
 
-const tokenizer = new markdoc.Tokenizer({ allowIndentation: true });
+const tokenizer = new markdoc.Tokenizer({
+  allowIndentation: true,
+  allowComments: true,
```
mfix-stripe, line 23: Feels to me like `allowComments` should be `true` by default, no?

src/parser.test.ts
```diff
@@ -636,4 +636,20 @@ describe('Markdown parser', function () {
...
+    this is a test
+
+    <!-- foo -->
+    `);
+
+    expect(example).toDeepEqualSubset({
+      type: 'document',
+      children: [
+        { type: 'paragraph' },
+        { type: 'comment', attributes: { content: 'foo' } },
```
mfix-stripe, line 651: Neat!

src/parser.ts
```diff
@@ -45,6 +45,7 @@ function handleAttrs(token: Token, type: string) {
     }
     case 'text':
     case 'code':
+    case 'comment':
```
mfix-stripe, line 48: Love how this part of the code is working.

src/tokenizer/index.ts
```diff
@@ -1,13 +1,17 @@
...
 import type Token from 'markdown-it/lib/token';
 
 export default class Tokenizer {
   private parser: MarkdownIt;
 
   constructor(
-    config: MarkdownIt.Options & { allowIndentation?: boolean } = {}
+    config: MarkdownIt.Options & {
+      allowIndentation?: boolean;
+      allowComments?: boolean;
```
mfix-stripe, line 13: Default to `true`?

src/tokenizer/plugins/comments.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      `);
+
+      expect(example).toDeepEqualSubset(output);
+    });
+
+    it('block comment one one line', function () {
+      const example = parse(`
+      this is a test
+
+      <!-- example comment -->
```
mfix-stripe, line 79: For these last few tests, should we include the `foo` after the comments just to ensure it keeps the content after a multi-line comment?

src/tokenizer/plugins/comments.ts
```diff
@@ -0,0 +1,50 @@
+import type MarkdownIt from 'markdown-it/lib';
+import type StateBlock from 'markdown-it/lib/rules_block/state_block';
+import type StateInline from 'markdown-it/lib/rules_inline/state_inline';
+
+const OPEN = '<!--';
```
mfix-stripe, line 5: I'm totally peaceful with this approach, but I wonder if we should use the same regex that markdown-it uses internally: https://github.com/markdown-it/markdown-it/blob/df4607f1d4d4be7fdc32e71c04109aea8cc373fa/lib/common/html_re.js#L18

```js
var comment     = '<!---->|<!--(?:-?[^>-])(?:-?[^-])*-->';
```

src/tokenizer/plugins/comments.ts
```diff
@@ -0,0 +1,50 @@
...
+function inline(state: StateInline, silent: boolean): boolean {
+  if (!state.src.startsWith(OPEN, state.pos)) return false;
+
+  const close = state.src.indexOf(CLOSE, state.pos);
+
+  if (!close) return false;
+  if (silent) return true;
+
+  const token = state.push('comment', '', 0);
+  token.content = state.src.slice(state.pos + OPEN.length, close).trim();
```
mfix-stripe, line 41:
```suggestion
  const content = state.src.slice(state.pos + OPEN.length, close);
  const token = state.push('comment', '', 0);
  token.content = content.trim();
```
Just to make it align with the function above, and highlight their differences.

mfix-stripe (commented):
This is great! Just have one question around whether we should make `allowComments` default `true`. PTAL @rpaul-stripe

spec/marktest/index.ts
mfix-stripe: Ahh. Makes sense.

mfix-stripe (approved):
Approving to unblock, but feel free to address the other comments and rerequest review 👍

src/tokenizer/plugins/comments.test.ts
rpaul-stripe: This was a good suggestion, adding it to the tests caught a bug. I'll have a push fixed shortly.

src/tokenizer/plugins/comments.ts
rpaul-stripe: I'd prefer to avoid using a regex here

mfix-stripe (approved):

PR #217, opened 2022-09-24
Remove console.error from formatter

mfix-stripe:
@rpaul-stripe WDYT about this? It feels weird to me to have this random side effect in the middle of the formatter. Should we yield the errors contents instead?

How do you want to approach this?


rpaul-stripe (approved):
Yes, I think ignoring error nodes in the formatter seems like the right approach to me. 👍

PR #222, opened 2022-09-28
Format titles for images and links

mfix-stripe:
PTAL @rpaul-stripe

### DIFF
--- a/src/formatter.test.ts
+++ b/src/formatter.test.ts
@@ -18,7 +18,8 @@ Markdoc is a **Markdown**-based \`syntax\` and _toolchain_ for creating ~~custom
 
 ---
 
-    ![Alt](/image)
+![Link](/href   "title")
+    ![Alt](/image   "title")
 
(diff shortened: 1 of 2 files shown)


matv-stripe (approved):

duncanma-stripe:
This looks a little odd, given that it has a Link and an image, but both using the image format (starting with a !), instead of 

```
[Link](/href "title")
![Alt](/image "title")
```

right? (I know it is just testing the formatter, but still )

mfix-stripe:
> This looks a little odd, given that it has a Link and an image, but both using the image format (starting with a !), instead of
> 
> ```
> [Link](/href "title")
> Alt
> ```
> 
> right? (I know it is just testing the formatter, but still )

First one is a link, second one is an image. I must be misunderstanding something @duncanma-stripe 🙂

matv-stripe:
@mfix-stripe 

in your diff, both of the examples are images (they both start with `!`)

mfix-stripe:
> @mfix-stripe
> 
> in your diff, both of the examples are images (they both start with `!`)

I see now! Yeah that was a typo. Fixing now.

PR #229, opened 2022-09-29
Add failing test case

mfix-stripe:


mfix-stripe:
Closing as I'm not sure we want to support this.

PR #233, opened 2022-09-29
Fix direct inline children within lists

mfix-stripe:


src/formatter.ts
```diff
@@ -134,20 +134,20 @@ function* formatNode(n: Node, o: Options = {}) {
...
     case 'paragraph': {
       yield NL;
-      yield indent;
       yield* formatChildren(n, no);
       yield* formatAnnotations(n);
       yield NL;
       break;
     }
     case 'inline': {
+      yield indent;
```
mfix-stripe, line 150: @rpaul-stripe what do you think about moving the `indent` from `paragraph` to `inline` here? Felt a little weird, but this solves for the case where there are `inline` elements in a `list` that don't have a `paragraph` for a parent. While still working for the case where there is a `paragraph`.

mfix-stripe (commented):
PTAL @rpaul-stripe

src/formatter.ts
rpaul-stripe: It does feel weird, but I guess it makes sense. I'm not opposed. 👍

rpaul-stripe (approved):

PR #237, opened 2022-10-04
Validation improvements

rpaul-stripe:
- Adds support for a `validate` function inside of an attribute schema
	- Optional property that allows performing arbitrary logic to validate the value of an attribute
- Adds a `validate` function for the global `id` attribute to make sure that the id starts with a letter
- Adds an `inline` schema property that indicates whether a tag should be used as a block or inline
	- When property is undefined, it allows it to be used as either block or inline
	- Adds validation rule that returns a "critical" error when the tag is used incorrectly
	- Adds an `inline` property value to the partial and table tags

### DIFF
--- a/src/transformer.ts
+++ b/src/transformer.ts
@@ -2,12 +2,27 @@ import Tag from './tag';
 import { Class } from './schema-types/class';
 import { isPromise } from './utils';
 import type { Config, Node, NodeType, Schema, Transformer } from './types';
+import { Value } from '..';
 
 type AttributesSchema = Schema['attributes'];
 
 export const globalAttributes: AttributesSchema = {
   class: { type: Class, render: true },
-  id: { type: String, render: true },
+  id: {
+    type: String,
+    render: true,
+    validate(value: Value, _config: Config) {
+      return typeof value === 'string' && value.match(/^[a-zA-Z]/)
+        ? []
+        : [
+            {
+              id: 'attribute-value-invalid',
+              level: 'error',
+              message: 'The id attribute must start with a letter',
+            },
+          ];
+    },
+  },
 };
 
 export default {
--- a/src/transformer.ts
+++ b/src/transformer.ts
@@ -18,7 +18,7 @@ export const globalAttributes: AttributesSchema = {
             {
               id: 'attribute-value-invalid',
               level: 'error',
-              message: 'The id attribute must start with a letter',
+              message: "The 'id' attribute must start with a letter",
             },
           ];
     },
--- a/src/transformer.ts
+++ b/src/transformer.ts
@@ -1,28 +1,14 @@
 import Tag from './tag';
 import { Class } from './schema-types/class';
+import { Id } from './schema-types/id';
 import { isPromise } from './utils';
 import type { Config, Node, NodeType, Schema, Transformer } from './types';
-import { Value } from '..';
 
 type AttributesSchema = Schema['attributes'];
 
 export const globalAttributes: AttributesSchema = {
   class: { type: Class, render: true },
-  id: {
-    type: String,
-    render: true,
-    validate(value: Value, _config: Config) {
-      return typeof value === 'string' && value.match(/^[a-zA-Z]/)
-        ? []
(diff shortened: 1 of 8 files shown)


src/transformer.ts
```diff
@@ -2,12 +2,27 @@ import Tag from './tag';
...
+    type: String,
+    render: true,
+    validate(value: Value, _config: Config) {
+      return typeof value === 'string' && value.match(/^[a-zA-Z]/)
+        ? []
+        : [
+            {
+              id: 'attribute-value-invalid',
+              level: 'error',
+              message: 'The id attribute must start with a letter',
```
mfix-stripe, line 21: I think this follows our error patterns better:
```suggestion
              message: "The 'id' attribute must start with a letter",
```

src/types.ts
```diff
@@ -114,6 +115,7 @@ export type SchemaAttribute = {
   required?: boolean;
   matches?: SchemaMatches | ((config: Config) => SchemaMatches);
   errorLevel?: ValidationError['level'];
+  validate?(value: Value, config: Config): ValidationError[];
```
mfix-stripe, line 118: Hmm. I feel like we should use use the `validate` function that exists within `type` instead of adding a new one here. Is there something this supports that custom attribute types doesn't?

src/transformer.ts
```diff
@@ -2,12 +2,27 @@ import Tag from './tag';
...
 import type { Config, Node, NodeType, Schema, Transformer } from './types';
+import { Value } from '..';
 
 type AttributesSchema = Schema['attributes'];
 
 export const globalAttributes: AttributesSchema = {
   class: { type: Class, render: true },
-  id: { type: String, render: true },
+  id: {
+    type: String,
```
mfix-stripe, line 12:
```suggestion
    type: ID,
```

And add this to `src/schema-types/class`:

```ts
export class ID implements CustomAttributeTypeInterface {
  validate(value: any): ValidationError[] {
    if (typeof value === 'string' && value.match(/^[a-zA-Z]/)) return [];

    return [
      {
        id: 'attribute-type-invalid',
        level: 'error',
        message: "The 'id' attribute must start with a letter",
      },
    ];
  }
}
```

src/validator.ts
```diff
@@ -211,6 +214,9 @@ export default function validator(node: Node, config: Config) {
       }
     }
 
+    if (typeof attrib.validate === 'function')
+      errors.push(...attrib.validate(value, config));
```
mfix-stripe, line 218: Remove this

src/transformer.ts
```diff
@@ -2,12 +2,27 @@ import Tag from './tag';
...
+    render: true,
+    validate(value: Value, _config: Config) {
+      return typeof value === 'string' && value.match(/^[a-zA-Z]/)
```
mfix-stripe, line 15: @rpaul-stripe is this true of all identifiers? Should we update `isIdentifier` so that they can't start with a number?

mfix-stripe (changes requested):
PTAL @rpaul-stripe

src/transformer.ts
rpaul-stripe: No, this is enforced only on the string value of the HTML `id` attribute, not all the things that are "identifiers" in the Markdoc tag grammar.

In the current implementation (before this PR) the validator already has a [hard-coded rule for ids](https://github.com/markdoc/markdoc/blob/main/src/validator.ts#L193) that prohibits starting them with a number. I went further in this PR and made it require a letter at the beginning in order to prevent people from making mistakes like e.g. `id="#foo"`. This is also consistent with the guidance on MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id#sect2

Granted, we may want to consider enforcing this rule for all Markdoc identifiers—but that would be a backwards-incompatible change since the grammar has always been permissive about this.

mfix-stripe: This makes sense, thanks!

mfix-stripe (approved):
This is fantastic!

PR #249, opened 2022-10-13
Fixes the type of the RenderableTreeNode

rpaul-stripe:
This PR tweaks the type of the RenderableTreeNode in order to reflect that other primitives besides `string` and `null` can be included in the output.

### DIFF
--- a/src/renderers/html.test.ts
+++ b/src/renderers/html.test.ts
@@ -12,29 +12,32 @@ function tag(
 
 describe('HTML renderer', function () {
   it('rendering a tag', function () {
-    const example = render(tag('h1', null, ['test'])).trim();
+    const example = render(tag('h1', undefined, ['test'])).trim();
     expect(example).toEqual('<h1>test</h1>');
   });
 
   it('rendering string child nodes', function () {
-    const example = tag('h1', null, ['test ', '1']);
+    const example = tag('h1', undefined, ['test ', '1']);
     expect(render(example)).toEqual('<h1>test 1</h1>');
   });
 
   it('rendering nested tags', function () {
-    const example = tag('div', null, [tag('p', null, ['test'])]);
+    const example = tag('div', undefined, [tag('p', undefined, ['test'])]);
 
     expect(render(example)).toEqual('<div><p>test</p></div>');
   });
 
   it('rendering parallel tags', function () {
-    const example = [tag('p', null, ['foo']), tag('p', null, ['bar'])];
+    const example = [
+      tag('p', undefined, ['foo']),
+      tag('p', undefined, ['bar']),
+    ];
 
     expect(render(example)).toEqual('<p>foo</p><p>bar</p>');
   });
 
   it('rendering a tag with an invalid child', function () {
-    const example = tag('div', null, ['test', { foo: 'bar' }]);
+    const example = tag('div', undefined, ['test', { foo: 'bar' }]);
 
     expect(render(example)).toEqual('<div>test</div>');
   });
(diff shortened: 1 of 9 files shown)


src/types.ts
```diff
@@ -93,7 +93,7 @@ export type NodeType =
 
 export type Primitive = null | boolean | number | string;
 
-export type RenderableTreeNode = Tag | string | null;
+export type RenderableTreeNode = Tag | Primitive;
```
matv-stripe, line 96: Technically, this is still not 100% accurate because `undefined` (or really anything) can be returned by the resolution of an `Ast.Variable` right?  Should we add `| object | undefined` here too then?

mfix-stripe: It we want to include `object`, then we should use `Tag | Scalar`. However, if we use `Tag | Scalar | undefined`, feels like we might just want to use `any`?

rpaul-stripe: I am happy to use `any`, as I think that is the most correct option. There were some objections to that in slack, however.

mfix-stripe: I think `any` is the most accurate.

adreyfus-stripe: `any` doesn't give us any type safety and wouldn't have prevented the incident though right? https://www.typescriptlang.org/docs/handbook/basic-types.html#any

mfix-stripe: I forgot this. Yeah let's do `Tag | Scalar` or `Tag | Scalar | undefined`

mfix-stripe: If we do `| undefined`, we should probably just make `undefined` part of `Primitive`.

mfix-stripe (commented):
I think we should do `Tag | Scalar` — PTAL @rpaul-stripe

rpaul-stripe:
Pushing a simpler version of this for now, will revisit the promise typing issues next week: https://github.com/markdoc/markdoc/pull/251

src/renderers/html.test.ts
```diff
@@ -1,39 +1,43 @@
...
   children: RenderableTreeNode[] = []
 ) {
-  return { name, attributes, children };
+  return new Tag(name, attributes, children);
 }
 
 describe('HTML renderer', function () {
   it('rendering a tag', function () {
-    const example = render(tag('h1', null, ['test'])).trim();
+    const example = render(tag('h1', undefined, ['test'])).trim();
```
matv-stripe, line 15: is intentional to have the `tag` helper function instead of using `Tag` class directly in the tests? (I was initially confused because I missed the function and was confused why we were passing `undefined`

src/tag.ts
```diff
@@ -6,6 +6,10 @@ export default class Tag<
 > {
   readonly $$mdtype = 'Tag' as const;
 
+  static isTag = (tag: any): tag is Tag => {
+    return '$$mdtype' in tag && tag.$$mdtype === 'Tag';
```
matv-stripe, line 10: should we add a test for this?

matv-stripe (approved):
2 small questions, looks good though

src/tag.ts
mfix-stripe: Tons of tests test for this.

src/renderers/html.test.ts
mfix-stripe: I think it's just to make the test cases smaller. We could replace with `new Tag` if we wanted.

mfix-stripe:
PTAL @rpaul-stripe

mfix-stripe (commented):
PTAL @rpaul-stripe

PR #265, opened 2022-10-22
fix transform return type

nazna:
Hello.

I love this useful library and recently tried an async transform. https://github.com/markdoc/markdoc/pull/109
Then I found what seems to be a small mistake in the type definition.
markdoc

I haven't read much of the code, but I am thinking that maybe I can fix this type.
I would be happy to check it out if you would like.

Thanks!

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -13,7 +13,7 @@ import transforms from './src/transforms';
 import { parseTags, isPromise } from './src/utils';
 import validator from './src/validator';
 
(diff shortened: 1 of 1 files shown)


nazna:
@mfix-stripe 
Hi.
Thanks for the comment.

I understand about the situation.
No problem at all!

I'll close this PR.
I'm looking forward to the experimental async transform API being officially supported 👍

PR #320, opened 2023-02-01
Fix null scalar formatting

emmatown:


src/formatter.test.ts
```diff
@@ -21,7 +21,7 @@ Markdoc is a **Markdown**-based \`syntax\` and _toolchain_ for creating ~~custom
 [Link](/href   "title")
     ![Alt](/image   "title")
 
-{% callout #id   .class  .class2   a="check" b={"e":{"with space": 5}} c=8 d=[1,    "2",true] %}
+{% callout #id   .class  .class2   a="check" b={"e":{"with space": 5}} c=8 d=[1,    "2",true,  null] %}
```
mfix-stripe, line 24: PTAL @rpaul-stripe — I didn't think we supported "null" in the syntax.

rpaul-stripe: We actually do support `null` as a valid scalar in Markdoc tags: https://github.com/markdoc/markdoc/blob/main/src/grammar/tag.pegjs#L119-L120

mfix-stripe: Ooh! Missed that in the grammar. Let's merge this then huh @rpaul-stripe ?

rpaul-stripe (approved):

mfix-stripe:
Thanks @emmatown!

Issue #10, opened 2022-05-11
Support for inline and block HTML content


marshall007:
The commonMark spec describes how [HTML blocks](https://spec.commonmark.org/0.30/#html-blocks) and [Raw HTML](https://spec.commonmark.org/0.30/#raw-html) should be treated:

> **4.6 HTML blocks**
>
> An [HTML block](https://spec.commonmark.org/0.30/#html-block) is a group of lines that is treated as raw HTML (and will not be escaped in HTML output).

>  **6.6 Raw HTML**
>
> Text between < and > that looks like an HTML tag is parsed as a raw HTML tag and will be rendered in HTML without escaping. Tag and attribute names are not limited to current HTML tags, so custom tags (and even, say, DocBook tags) may be used.

Example:

```md
# Test <em>Emphasis</em>

<a href="http://github.com/markdoc/markdoc">GitHub</a>
```

Are there plans to support this aspect of the spec?

If so, it's worth noting that other markdown implementations I've seen tend to follow the spec literally and do not parse HTML content in any way. In other words, raw HTML content is just treated as a string in the AST. The downside to this approach is that it prevents you from introspecting raw HTML content.

For example, if you wish to write a validator that ensures the integrity of links on a page you don't really care whether the links are authored natively in markdown or as raw `<a>` tags. Likewise, when generating a table of contents, you want to generate IDs for and include all header tags.

Cheers, and congrats on the first release!

rpaul-stripe:
The `markdown-it` library has an option (`html: true`) that can be enabled to get it to identify HTML content in Markdown according to the rules in the CommonMark specification. When this option is passed into the Markdoc tokenizer, a document with HTML content will fail to parse because Markdoc doesn't define a corresponding HTML node type.

It's fairly trivial to add an HTML node type to `schema.ts` so that we can capture these HTML strings and expose them in the Markdoc AST. We can add a new renderable tree type that is specifically for pass-through HTML content, and we can have the HTML node type output that during Markdoc's `transform` phase.

It's straightforward to support this in the HTML renderer, because we can just append the content to the output string when we encounter it in the renderable tree. The problem, however, is figuring out a good way to support this in the React renderer.

[...]

PR #344, opened 2023-03-14
Adds support for arbitrary HTML content in Markdoc

rpaul-stripe:
Markdoc currently treats HTML as though it is plain text, automatically replacing the syntactically-relevant characters with equivalent entity references. This PR introduces a new optional feature that enables the use of raw HTML content inside of a Markdoc document.

For a full description of the feature, refer to the [RFP](https://github.com/markdoc/markdoc/discussions/343).

- Adds a new render tree AST type called `Raw` that is used to represent raw HTML content in a render tree
- Tweaks the Markdoc parser so that it recognizes the `html_block` and `html_inline` tokens from `markdown-it`
- Modifies the HTML and React renderers to incorporate support for handling the `Raw` render tree node type
- Introduces a new default schema definition for the `html` document type
- Adds support for customizing the tokenizer settings per test in Marktest
- Adds test cases that evaluate whether inline and block raw HTML work as expected


src/ast/raw.ts
```diff
@@ -0,0 +1,15 @@
...
+  readonly $$mdtype = 'Raw' as const;
+
+  static isRaw = (node: any): node is Raw => {
+    return !!(node?.$$mdtype === 'Raw');
+  };
+
+  content: string;
+  inline: boolean;
+
+  constructor(content: string, inline: boolean) {
```
mfix-stripe, line 11: I think I might be missing something, but is there any reason we can't treat HTML as a scalar of some sort? And not have to introduce this new AST type?

mfix-stripe (commented):
I am really nervous about adding this capability to Markdoc. 

I know it is opt-in, but I feel like it _irreversibly_ complicates the Markdoc syntax, while expanding the surface area of our APIs quite a bit (e.g. `raw?: (content: string, inline: boolean) => ReactNode`). All without adding too much end-user capability (everything you can do with HTML you can do with Tags)

IMO, we should avoid merging this, unless we get strong consensus from the community in the [RFP](https://github.com/markdoc/markdoc/discussions/343).

If you feel strongly we should support this, I can help unblock, but I do wonder if we can do so in a way without complicating our renderers or AST too much. Like, could we do so by treating the HTML content as a scalar of some sort (string)? Might be missing something here.

src/ast/raw.ts
rpaul-stripe: Conventional strings are [escaped](https://github.com/markdoc/markdoc/blob/115699172cbb824a6a8b39e3edec0a868e596aeb/src/renderers/html.ts#L26-L27) for security reasons—symbols like '<' and '>' are replaced with the equivalent HTML entity references. If we pass the raw HTML all the way through as a string, it's going to render literally instead of as HTML. Putting the raw content in a dedicated AST node makes it possible to distinguish it from regular text.

It's also important to capture and preserve whether it originated in a block or inline context because that may have implications for how you render it (e.g. if you are going to put it in a `dangerouslySetInnerHtml` attribute on an HTML element, you might want to use `div` or `span` depending on where the content originated)

rpaul-stripe:
> I am really nervous about adding this capability to Markdoc. ...

I hear you. I feel strongly that we should support this on the parsing side. On the rendering side, I still have some reservations.

CommonMark-compliant HTML parsing is fully supported by markdown-it and is technically already exposed in Markdoc—if a user passes the `html: true` Tokenizer option into Markdoc today, markdown-it will tokenize the content, but then the Markdoc parser will choke on the resulting tokens because it doesn't recognize them. At minimum, I think we absolutely need to add the `html` node type and support the markdown-it `html_block` and `html_inline` tokens in the parser so that this behaves predictably and doesn't create unpleasant surprises for users.

The `Raw` render tree node is a very powerful primitive, and it's worth noting that beyond being used for HTML passthrough it can also be used in other custom tag transform functions to support a range of additional use cases where you might want literal, unprocessed output. I actually had a pretty similar feature in the original Markdoc prototype. I think that it is inherently useful, especially for people who are working with custom renderers. But it's also potentially pretty dangerous for users who don't understand the potential security footguns.

On the rendering side, my big concern is with React's lack of native support for handling interleaved raw markup. The fact that React doesn't have a well-paved path for doing this is what necessitates the janky callback hack, and I agree that the callback is pretty undesirable. Rendering is also modular enough that we could potentially omit this feature from core and supply an alternate renderer implementation that includes it for people who really want this. That would also make it possible to integrate more tightly with React-specific stuff like the interweaved library.

What do you think about keeping it in the parser, making it generate a `Raw` node, supporting it in the HTML renderer, but then not supporting it ootb in the React renderer?

src/schema.ts
```diff
@@ -239,5 +242,17 @@ export const comment = {
...
 
+export const html: Schema = {
+  attributes: {
+    content: { type: String, required: true },
+  },
+
+  transform(node) {
+    return node.attributes.content
+      ? new Raw(node.attributes.content, node.inline)
+      : null;
```
mfix-stripe, line 253: Do we need this ternary?

mfix-stripe:
Thanks for the thorough explanation @rpaul-stripe. I am totally peaceful with this plan: 
> What do you think about keeping it in the parser, making it generate a Raw node, supporting it in the HTML renderer, but then not supporting it ootb in the React renderer?

I do wonder if we should call the `Raw` node something that indicates the security danger you expressed above. `DangerousRaw` or something? I kinda don't like it, and don't have strong feelings, so just wanted to put the question out there.

alex-sherwin:
I had to come up with a workaround for this issue (the suggestions in #10 are too naive) to support any arbitrary tree/nesting of HTML and Markdoc tags.

I'm building something which takes the output of an existing documentation platform whose markup is a mix of raw HTML and custom macro's, and I'm converting en-masse to Markdoc using the raw HTML + 1:1 Markdoc tag's for the macro equivalents implemented as React components.

My solution introduces a "html" Markdoc tag and at tag processing time uses https://www.npmjs.com/package/htmlparser2 (which is an event based lazy/incremental HTML parser, which is important) to build a new tree of Markdoc tag's which is a tree of native react HTML nodes (e.g. `p`, `div`, etc) and any nested other Markdoc tags.

Simple example (indentation only for clarity sake):

```markdown
{% html %}
  <p>outer1</p>
  <p>outer2</p>
  {% callout icon="🔥" %}
    {% html %}
    <div>inner</div>
    {% /html %}
  {% /callout %}
{% /html %}
```

Which will when processed by the **html** tag will produce a RenderableTreeNode tree like this:

```
Tag("p")
    children: ["outer1"]
Tag("p")
    children: ["outer2"]
Tag("Callout")
    children: [
        Tag("div")
            children: ["inner"]
    ]
```

And the React rendering component is a simple pass-through (just renders the given children, no wrappers)

I don't love that I need to have an `{% html %}` tag around every area of HTML (i.e. under the Callout tag it needs another `{% html %}` to parse and HTML below that

However, it should work for any arbitrary complexity.

If you're interested to see my implementation I can share it.

alex-sherwin:
The new `Raw` TypeScript type does not appear to be exported in the final package build

alex-sherwin:
The new `html` node is not added to the `NodeType` union type in src/types.ts

alex-sherwin:
I've been testing this branch... unless I'm mis-understanding, it seems pretty rough to use in the real-world, mixing HTML content with Markdoc tags, I think, doesn't result in a usable sequence of events/rawHandler callback's.

The **rawHandler** callback is simply invoked every time some sequence of HTML is encountered (which could be, for example, just a random close tag), and is expected to produce a **React.ReactNode** (which must know about it's children **at this time**, which might need to be something like another Markdoc Tag that follows the HTML open tags the **rawHandler** just processed (and thus won't be able to render the children properly), for example this wouldn't work:

```markdown
<div>

{% callout %}
content
{% /callout %}

</div>
```

Because **rawHandler** would be called with string `<div>`, and you could render a native div at this time, but you'd also need to render it's children (which should be the callout tag), but you won't be able to.

And the approach of customizing the new `html` node's `transform` function largely suffers the same fate (you are only presented a given fragment of HTML at a time, without any context of any other Markdoc tags that follow in the markup that should actually be children of the currently-being-transformed HTML node in the final HTML DOM tree).

This level of hooks would make it hard (impossible?) to build a proper set of React components that represents the HTML tree.

Maybe I'm missing something fundamental, but the the RFP and issue #10 don't seem to consider an approach that utilizes the fact that React has first-class support for HTML tags in JSX syntax, e.g. the typical:

```jsx
function MyReactComponent() {
  return <p>im a paragraph</p>;
}
```

Markdoc could, in theory, map the HTML input from the original markup into these tags, I'm already doing this in my custom html Markdoc tag, e.g.:

```typescript
const htmlNodeTag = new Tag<keyof JSX.IntrinsicElements, {}>(
 "p",
  { class: "css classes", id: "fragment-id" },
  [],
);
```

So if Markdoc is able to tokenize HTML input to detect whenever an HTML tag opens, it can just build a tree of Markdoc **Tag** class instances that mirror this HTML tree, and place other non-HTML Markdoc Tags in this same tree.

I've accomplished this in a custom Markdoc html tag, but it requires me to live with the restrictions of being a custom tag, so I have to put `{% html %}` markup around HTML content.

I would prefer if this PR/html feature would provide a first-class way to accomplish the same (let me plug in **htmlparser2** at the appropriate tag transform time) without having to wrap HTML content with a completely custom Markdoc tag..

Here is the custom **html** Tag I created which supports any arbitrary nesting/mixing of HTML and Markdoc tags, it utilizes the *htmlparser2* package as a lazy/incremental HTML DOM parser and is able to build a new Tag tree that mixes HTML DOM nodes and other custom Markdoc Tag nodes

(you can assume `parseInlineCSSToReactObject` to be black box that turns inline CSS style strings into react-compatible CSS objects)

This approach (I think) works for any arbitrary complexity and is also fully SSR/RSC capable (including the handling of inline **style** HTML attributes)

```typescript
// third party
import { Tag, Config, RenderableTreeNode, Schema } from "@markdoc/markdoc";
import { Parser } from "htmlparser2";

// local
import { parseInlineCSSToReactObject } from "../../../../util/parse-inline-css-to-react";

export const html: Schema<Config, "Html"> = {
  render: "Html",
  transform: (node, config) => {

    // console.log("Html");

    /**
     * markdoc (as of 0.2.2, see https://github.com/markdoc/markdoc/issues/10)
     * 
     * does not currently properly support inline/block native HTML markup (but, it should, as CommonMark does)
     * 
     * the result of node.transformChildren(..) will be a tree of markdoc Tag instances which is a mix of "p" tags (HTML paragraph DOM nodes)
     * whose text contents are the actual HTML we need to render, which are interleaved with any other Markdoc Tags that were in the original markup.
     * 
     * we cannot simply concatenate all the HTML together and parse it and/or set it using React's dangerouslySetInnerHTML because:
     *   1. using dangerouslySetInnerHTML requires being set on a wrapper DOM element, which means we have to introduce a HTML DOM node that wasn't in the original markup to begin with.
     *   2. the HTML markup may be interleaved with other Markdoc tags at arbitrarily complex levels of nesting
     *   3. inline CSS styles (React native DOM elements can't set "style" as a simple string, the only option is to parse any inline styles into a React-compatible CSS.Properties object)
     * 
     * the solution is to take this tree, traverse it, and build a new tree while incrementally parsing HTML strings as they are visited, and interleaving other Markdoc Tag's as they
     * are visited.
     */
    const unsafeChildrenRenderableTrees = node.transformChildren(config);

    // console.log("unsafeChildrenRenderableTrees", unsafeChildrenRenderableTrees);

    // incrementally parse HTML, build new children Tag tree that matches the parsed HTML while interleaving any other Markdoc tags
    const transformedChildRenderableTrees = transformMixedHTMLAndMarkdocTree(unsafeChildrenRenderableTrees);

    /**
     * the new Tag tree generated is a mixed array of React HTML DOM elements (parsed HTML) and other Markdoc Tags.
     * 
     * they are ready to be rendered as-is, so send them through a simple "passthrough" React component which just renders the children under a fragment without any processing
     */
    return new Tag<"Passthrough", {}>(
      "Passthrough",
      {},
      transformedChildRenderableTrees
    );

  },
};


/**
 * we don't actually have to traverse a tree here, we only have to deal with the current flat array of RenderableTreeNode[] (markdoc Tags)
 * 
 * the output, however, will trees of markdoc Tag's that match the parsed HTML DOM string(s) interleaved with any other Markdoc Tag's encountered in the markup
 */
const transformMixedHTMLAndMarkdocTree = (nodes: RenderableTreeNode[]): RenderableTreeNode[] => {

  // buffer of currently un-processed raw HTML string data, lazily build it up until it requires processing (once we visit a non-HTML Tag)
  let unprocessedHtmlBuffer = "";

  // holds the root RenderableTreeNode's for the result
  const resultRootTags: RenderableTreeNode[] = [];

  // track the current stack of HTML Tag's
  let currentHtmlTagStack: Tag[] = [];

  // create a incremental HTML parser instance
  const parser = new Parser({

    // HTML Tag opens
    onopentag(htmlTagName: keyof JSX.IntrinsicElements, rawHtmlAttributes: Record<string, string | undefined>) {

      // "style" is the only [known] problematic HTML attribute, it needs to be processed from a raw CSS string into a React-compatible CSS.Properties object
      const { style, ...safeHtmlNodeAttributes } = rawHtmlAttributes;

      const reactHtmlAttributes = safeHtmlNodeAttributes as React.HTMLAttributes<unknown>;

      // if "style" was set, process into a CSS.Properties compatible object
      if (style) {
        const styleObject = parseInlineCSSToReactObject(style);
        reactHtmlAttributes.style = styleObject;
      }

      // if (htmlTagName === "p") {
      console.log(`+ ${htmlTagName}`);
      // console.log(`current tag`, currentHtmlTagStack.at(-1));
      // }

      // create a React HTML DOM markdoc Tag (e.g. "p", "div", etc.)
      const htmlNodeTag = new Tag<keyof JSX.IntrinsicElements, {}>(
        htmlTagName,
        reactHtmlAttributes,
        [],
      );

      if (currentHtmlTagStack.length === 0) {
        // this is a new root node, track it for the overall result
        resultRootTags.push(htmlNodeTag);
      } else {
        // not a root, add it as a child children of the current node
        const parent = currentHtmlTagStack.at(-1)!;
        parent.children.push(htmlNodeTag);
      }

      // push on the stack as the current node
      currentHtmlTagStack.push(htmlNodeTag);

    },

    // either raw text (not inside an element) or text under a HTML DOM node encountered
    ontext(text) {
      if (typeof text === "string" && text.trim().length === 0) {
        return;
      }
      // console.log(`got text [${text}]`);
      // check the current HTML tag
      const current = currentHtmlTagStack.at(-1);
      if (current) {
        // append the text to the current html tags children
        current.children.push(text);
      } else {
        // no tag exists to hold it, push a new root RenderableTreeNode scalar value
        resultRootTags.push(text);
      }
    },

    // HTML tag closes
    onclosetag(htmlTagName: string) {
      // if (htmlTagName === "p") {
      console.log(`- ${htmlTagName}`);
      // }

      currentHtmlTagStack.pop();
    },

  });

  // function which takes the current unprocessed HTML chunk and pushes it into the parser
  const processBufferedHTML = () => {
    if (unprocessedHtmlBuffer.trim().length > 0) {
      parser.write(unprocessedHtmlBuffer);
      // reset the current html buffer
      unprocessedHtmlBuffer = "";
    }
  };

  // for each RenderableTreeNode
  for (const node of nodes) {

    // it definitely should be a Tag instance (else Error)
    if (node instanceof Tag) {

      // this is a wrapper for raw HTML strings, concat consecutive occurrences as a 'chunk' to be parsed later
      if (node.name === "p") {
        if (Array.isArray(node.children)) {
          for (const child of node.children) {
            if (typeof child === "string") {
              unprocessedHtmlBuffer += child;
            } else {
              // console.error("found non-string child of a <p>!", child);
              // throw new Error("found non-string child of a <p>!");

              processBufferedHTML();

              // now, add the current non-HTML Markdoc Tag
              const currentTag = currentHtmlTagStack.at(-1);
              if (currentTag) {
                // there is a current tag tree, add this Markdoc Tag to the current nodes children
                currentTag.children.push(child);
              } else {
                // no current root tag active, push it to the root RenderableTreeNode array
                resultRootTags.push(child);
              }
            }
          }
        }
      } else {

        // some other Markdoc tag (non-HTML), it needs to be inserted into the current result tree(s)

        // first, if we have buffered HTML, processig it now to maintain our output Tag tree to match all the HTML visited so far
        processBufferedHTML();

        // now, add the current non-HTML Markdoc Tag
        const currentTag = currentHtmlTagStack.at(-1);
        if (currentTag) {
          // there is a current tag tree, add this Markdoc Tag to the current nodes children
          currentTag.children.push(node);
        } else {
          // no current root tag active, push it to the root RenderableTreeNode array
          resultRootTags.push(node);
        }

      }
    } else {
      throw new Error("not a Tag class instance, this is not expected and must be handled");
    }
  }

  // process any remaining buffered HTML
  processBufferedHTML();

  return resultRootTags;
};
```

Some example complex'ish markup this can process:

```markdown
abc
{% html %}

before

<p style="color: hotpink; background-color: yellow">hi1a</p>
<p>hi1b</p>
<p>hi1c</p>

<p id="123">hi2</p>

<div class="wrap-tag"><div class="wrap2">
<p>hi3</p>
{% callout icon="🔥" %}
stuff
{% html %}
<div class="inner-wrap">
<span style="background-color: hotpink; color: white">
callout content</span>
<p style="box-shadow: rgba(60, 64, 67, 0.3) 0px 1px 2px 0px, rgba(60, 64, 67, 0.15) 0px 2px 6px 2px;">with <br/>box<br/>shadow!</p>

</div>
{% /html %}
{% /callout %}

</div>
<p>hi4</p>
</div>

<br />
<br/>

{% /html %}
```

alex-sherwin:
Sorry for the long replies, but this is a big feature for me (and others) and I'm trying to figure out how it'll be workable

My previous comment still stands, I don't think simply augmenting the **html** node behavior or implementing a **rawHandler** on the React renderer is workable.

However, the changes from this PR result in a better AST with the information needed at that level (so after **markdoc.parse** but before **markdoc.transform**).

I adapted my algorithm from the previous comment to be a recursive AST mutation to achieve the result I want (parse the HTML into a tree, interleave other Markdoc tags and render as intrinsic JSX DOM via React)

## I needed to make only a few minor changes to the markdoc lib to enable this:

1. Export the `Node` class (currently only the `Node` **type** is exported) -- So that I can mutate the AST and instantiate new `Node` class instances
2. Add the new *html* node type to the `NodeType` union type (not strictly necessary but makes my code type safe)

## With only those changes, one can do the following:

1. Override the default *html* node transform behavior in the markdoc config:
    ```typescript
    nodes: {
      html: {
        attributes: {
          // denotes whether the AST mutate function as already processed a given 'html' node
          processed: {
            type: Boolean,
            required: false,
          },
          htmlTagName: {
            type: String,
            required: true,
          },
          htmlAttributes: {
            type: Object,
            required: false,
          }
        },
        transform: function (node, config) {
          const children = node.transformChildren(config);
          // take the 'html' node attribute metadata and create a Tag to render an JSX Intrinsic HTML DOM element
          // directly as a React component with all HTML DOM element attributes
          const renderable = new Tag(node.attributes.htmlTagName, node.attributes.htmlAttributes, children);
          return renderable;
        }
      }
    },
    ```
2. After **markdoc.parse**, take the resulting AST `Node` (type "document") and mutate it with a recursive algorithm that will incrementally/lazily parse raw HTML and build a matching tree of `Node(type=html)` instances as they are parsed, while interleaving other non-HTML Markdoc `Node` instances in the resulting tree at appropriate locations.  This is my AST mutate function:
    ```typescript
    import { Node } from "@markdoc/markdoc";
    import { Parser } from "htmlparser2";
    
    // local
    import { parseInlineCSSToReactObject } from "../../util/parse-inline-css-to-react";
    
    
    export default function htmlAstTransform(originalNode: Node): void {
    
      // buffer of currently un-processed raw HTML string data, lazily build it up until it requires processing (once we visit a non-HTML node)
      let unprocessedHtmlBuffer = "";
    
      // holds the root Node's for the result
      const resultingChildren: Node[] = [];
    
      // track the current stack of HTML Tag's
      let currentHtmlNodeStack: Node[] = [];
    
      // create a incremental HTML parser instance
      const parser = new Parser({
    
        onopentag(htmlTagName: keyof JSX.IntrinsicElements, rawHtmlAttributes: Record<string, string | undefined>) {
    
          // pull out a "style" attribute (special attribute that needs processing)
          const { style, ...safeHtmlNodeAttributes } = rawHtmlAttributes;
    
          // re-type
          const reactHtmlAttributes = safeHtmlNodeAttributes as React.HTMLAttributes<unknown>;
    
          // if style was set, parse it into a React compatible React.CSSProperties object
          if (style) {
            const styleObject = parseInlineCSSToReactObject(style);
            reactHtmlAttributes.style = styleObject;
          }
    
          // create a Markdoc "html" node which holds the metadata we need to render as a native JSX HTML DOM element
          const htmlNodeTag = new Node(
            "html",
            { processed: true, htmlTagName: htmlTagName, htmlAttributes: reactHtmlAttributes },
            [],
          );
    
          if (currentHtmlNodeStack.length === 0) {
            // this is a new root node, track it at the root of the children
            resultingChildren.push(htmlNodeTag);
          } else {
            // not a root, add it as a child of the current HTML node
            const parent = currentHtmlNodeStack.at(-1)!;
            parent.children.push(htmlNodeTag);
          }
    
          // push on the stack as the current node
          currentHtmlNodeStack.push(htmlNodeTag);
    
        },
    
        // either raw text (not inside an element) or text under a HTML DOM node encountered
        ontext(text) {
          if (typeof text === "string" && text.trim().length === 0) {
            return;
          }
    
          // build a simple "text" node
          const textNode = new Node("text", { content: text });
    
          // check the current HTML tag
          const current = currentHtmlNodeStack.at(-1);
    
          if (current) {
            // append the text to the current html tags children
            current.children.push(textNode);
          } else {
            // no tag exists to hold it, push a new root RenderableTreeNode scalar value
            resultingChildren.push(textNode);
          }
        },
    
        // HTML tag closes
        onclosetag(htmlTagName: string) {
          currentHtmlNodeStack.pop();
        },
    
      });
    
      // takes the current unprocessed HTML chunk and pushes it into the parser
      const processBufferedHTML = () => {
        if (unprocessedHtmlBuffer.trim().length > 0) {
          parser.write(unprocessedHtmlBuffer);
          // reset the current html buffer
          unprocessedHtmlBuffer = "";
        }
      };
    
      // for each child Node
      for (const node of originalNode.children) {
    
        // if this is a "html" node from the markdown-it processing (our processed flag is not set)
        if (node.type === "html" && node.attributes.processed !== true) {
          // buffer the raw HTML to be parsed when needed
          unprocessedHtmlBuffer += node.attributes.content;
        } else {
    
          // some other Markdoc node (non-HTML), it needs to be inserted into the current result node tree position
    
          // first, if we have buffered HTML, processig it now to maintain our output Node tree to match all the HTML at this level so far
          processBufferedHTML();
    
          // now, add the current non-HTML Markdoc node
          const currentTag = currentHtmlNodeStack.at(-1);
          if (currentTag) {
            // there is a current Node tree, add this Markdoc Tag to the current nodes children
            currentTag.children.push(node);
          } else {
            // no current root Node active, push it to the root children nodes array
            resultingChildren.push(node);
          }
    
        }
    
      }
    
      // process any remaining buffered HTML
      processBufferedHTML();
    
      originalNode.children = resultingChildren;
    
      // recurse
      for (const node of originalNode.children) {
        htmlAstTransform(node);
      }
    
    };
    ```

With that, you get full HTML support with arbitrary nesting complexity mixed with other Markdoc tags in the markup

## Sample complex markup

```markdown

<p id="p1">paragraph one</p>

markdown with inline <span id="inline-span-1" class="a b c" style="color: hotpink;">inline-span</span>

{% callout icon="🔥" %}
<span id="callout-span1">root callout!!</span>
{% /callout %}


<div style="background-color: hotpink; color: white">
  some div text mixed with
  <span id="a-span"> a span</span>
  <p style="color: yellow;">a div paragraph!</p>
  <div id="nested_div">

{% callout icon="🔥" %}
text before html div
<div id="nested-call-out-div" style="background-color: blue; color: white;">
<p id="inner-callout-p">nested callout inside paragraph</p>
text inside div
</div>
text after html div

{% /callout %}

  </div>
</div>
```

## Sample rendered DOM

```html
<article>
  <p id="p1">paragraph one</p>
  <p>markdown with inline <span id="inline-span-1" style="color:hotpink" class="a b c">inline-span</span></p>
  <div class="callout">
    <div class="content">
      <div><span>icon: <!-- -->🔥</span></div>
      <div class="copy"><span class="title"></span>
        <p><span id="callout-span1">root callout!!</span></p><span>!!!! 22 !!!</span>
      </div>
    </div>
  </div>
  <div style="background-color:hotpink;color:white">
    some div text mixed with
    <span id="a-span"> a span</span>
    <p style="color:yellow">a div paragraph!</p>
    <div id="nested_div">
      <div class="callout">
        <div class="content">
          <div><span>icon: <!-- -->🔥</span></div>
          <div class="copy"><span class="title"></span>
            <p>text before html div</p>
            <div id="nested-call-out-div" style="background-color:blue;color:white">
              <p id="inner-callout-p">nested callout inside paragraph</p>
              text inside div
            </div>
            text after html div
            <span>!!!! 22 !!!</span>
          </div>
        </div>
      </div>
    </div>
  </div>
</article>
```

## Summary

So in summary, if we could also simply

1. Export the `Node` class to enable one to build their own AST mutation function(s)
2. Fix the `NodeType` type union to include the new `html` node type

End-users of the library can utilize the improved AST this PR provides (makes the HTML detection much better in the resulting AST versus before this PR), and provide their own AST mutator (like demonstrated) to get any level of HTML rendering needed

src/schema.ts
rpaul-stripe: Eh, we probably don't. I don't think there's a case where the parser actually returns any empty html token, so this may have been overly defensive.

rpaul-stripe:
@alex-sherwin 

> Export the Node class (currently only the Node type is exported) -- So that I can mutate the AST and instantiate new Node class instances

I think the `Node` type is currently exported incidentally from the top-level because of `export * from ./src/types`, but the actual class is accessible via the top-level `Ast` export, e.g.: `Markdoc.Ast.Node`. So I didn't change anything there, but let me know if that doesn't work for you.

I added the html node to the `NodeType` union, thanks for catching that.

alex-sherwin:
Hi @rpaul-stripe 

Thanks for reviewing my lengthy complaints :)

* I like your working example at token parsing time, appears to be much simpler in nature.  I will check out if it suits my needs fully and probably switch to this approach
* Yes somehow I missed the Node type being available under Ast, sorry, that works for me .. I've got my example AST mutator function (from above) working 100% from the current tip of this branch/PR

Thank you very much, I'm personally happy with the state of this and find it workable for my needs

rpaul-stripe:
@mfix-stripe

- Removed the callback from the React renderers and made it so that they convert the Raw render tree node to `null`
- Renamed `Raw` to `UnsafeRaw` (I liked "unsafe" better than "dangerous" — subtle semantic distinction, but seems more correct. We don't actually know that it is dangerous for any given content, we just know that it isn't safe).
- Removed the unnecessary ternary in the node transform

alex-sherwin:
One last thing @rpaul-stripe that came up when trying to use your working example to do the transform on the tokens.

Currently the `Token` type used by markdown-it is not exported from this library, so it's not present in any of the typings that are built into the dist.

But you're bundling markdown-it inside your dist, so, in order for me to get typings (as of now) I'd need to have an explicit dependency on **@types/markdown-it**

If that's what you intend, that's fine.. or, since you're bundling markdown-it inside markdoc, should the src/types.ts file re-export it?

rpaul-stripe:
@alex-sherwin I added that type to the exports. 👍

src/types.ts
```diff
@@ -1,9 +1,11 @@
 import type Func from './ast/function';
 import type Node from './ast/node';
 import type Var from './ast/variable';
+import type UnsafeRaw from './raw';
 import type Tag from './tag';
+import type Token from 'markdown-it/lib/token';
 
-export type { Node, Tag };
+export type { Node, Tag, Token };
```
mfix-stripe, line 8: @rpaul-stripe are we sure we want to expose `Token` here? Just want to be safe around over-coupling us to `markdown-it`

mfix-stripe (commented):
This looks good to me! Want to make sure the comments in the PR thread are resolved before we merge this though.

src/types.ts
alex-sherwin: While it's convenient since you're bundling markdown-it internally, that seems like a valid point, I don't have a problem with using a direct dependency on @types/markdown-it instead.

vital4u4ok:
@rpaul-stripe Hi, do you have any estimates on merging this PR?

rpaul-stripe:
We have decided not to move forward with this PR. Instead, we encourage users who want native HTML support in Markdoc to rely on a token transform that converts the HTML content into Markdoc tags. For more details, see the [note in the related issue](https://github.com/markdoc/markdoc/issues/10#issuecomment-1492560830).

alex-sherwin:
I think this sounds like the right thing to do after the transform approaches that we came up with

Any chance we could get the minor type info changes that were rolled into this PR, though?

Namely:

1. Export the Node class to enable one to build their own AST mutation function(s)
1. Fix the NodeType type union to include the new html node type

rpaul-stripe:
@alex-sherwin the Node class is currently exported on the `Ast` object that is at the top level of the module, e.g.:

```
import Markdoc from "@markdoc/markdoc"

const node = new Markdoc.Ast.Node(...)
```

alex-sherwin:
Thanks @rpaul-stripe 

This was a very constructive issue/PR feedback loop in the end :)

philippaz272006-commits (commented):
****

PR #348, opened 2023-03-28
Preserve list marker

rpaul-stripe:
This PR adds an attribute called `marker` to `list` nodes that preserves the marker symbol used for the list in the source document. For ordered lists, this is either `)` or `.` and for unordered lists it is either `-` or `*`. The marker is already exposed by the tokenizer, so the parser just has to take this value and assign it to the attribute.

This PR also modifies the formatter so that it uses the marker provided by the source content when available instead of defaulting to a specific value. Previously, the formatter was programmed to always use `1.` for ordered lists and `-` for unordered lists in order to enforce consistency.

The problem that has arisen with this behavior is that in cases where there are two separate lists that appear in immediate succession in a document, the marker needs to be different in order for the parser to correctly identify them as separate lists. This behavior is demonstrated by [Example 301](https://spec.commonmark.org/0.30/#example-301) in the CommonMark specification. Making the formatter respect the original marker used in the source content ensures that it will not erroneously combine separate lists that appear in sequence.

### DIFF
--- a/src/formatter.test.ts
+++ b/src/formatter.test.ts
@@ -445,6 +445,41 @@ Yes!
     check(source, expected);
     stable(expected);
   });
+
+  it('preserving list marker', () => {
+    const source = `
+- foo
+- bar
+* baz
+* qux
+
+
+1) foo
+2) bar
+3) baz
+1. foo
+2. bar
+3. baz
+`;
+    const expected = `
+- foo
+- bar
+
+* baz
+* qux
+
+1) foo
+1) bar
+1) baz
+
+1. foo
+1. bar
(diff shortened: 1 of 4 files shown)


src/formatter.test.ts
```diff
@@ -445,6 +445,41 @@ Yes!
...
+
+  it('preserving list marker', () => {
+    const source = `
+- foo
+- bar
+* baz
+* qux
+
+
+1) foo
```
mfix-stripe, line 457: Did not know `1)` works in Markdown

mfix-stripe (approved):

Issue #369, opened 2023-05-09
Emphasis written using `*` without leading and trailing whitespace is incorrectly formatted to use `_`


emmatown:
### What happened?

For a Markdoc document like this:

```md
a*b*c
```

The formatter turns it into this:

```md
a_b_c
```

This formatting isn't correct because the underscores won't be interpreted as emphasis, they're just interpreted as text. https://spec.commonmark.org/0.30/#emphasis-and-strong-emphasis explains how `_` vs `*` are different.

### To reproduce

1. Create a markdoc document like above
2. Run the formatter on it

### Version

0.2.2

rpaul-stripe:
It looks like markdown-it captures the actual syntax used for emphasis in a `markup` attribute on the token. I'm going to look at preserving that in the AST so we can hopefully reproduce it exactly in the formatter.

PR #373, opened 2023-05-09
Preserves em and strong markers in the AST

rpaul-stripe:
This PR addresses issue #369. It makes the parser capture the actual symbols used for `em` and `strong` nodes and preserve that information in the AST so that it can be used in the formatter to reproduce the content accurately.

- Parser modifications
  - Modifies the parser to capture the emphasis markup for `em` and `strong` nodes and apply it to a node attribute called `marker`
  - Adds a Jasmine test case to ensure that emphasis nodes get the proper `marker` attribute
- Schema modifications
  - Adds a non-rendering `marker` attribute to the `strong` and `em` node schemas
- Formatter modifications
  - Modifies the formatter so that it renders `em `and `strong` nodes using the original emphasis marker from the source content and falls back on `*` or `**` when not available
  - Adds a Jasmine test case to the formatter to ensure that the output matches the source

Closes https://github.com/markdoc/markdoc/issues/369

### DIFF
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -305,15 +305,15 @@ function* formatNode(n: Node, o: Options = {}) {
       break;
     }
     case 'strong': {
-      yield '**';
+      yield n.attributes.marker ?? '**';
       yield* formatInline(formatChildren(n, no));
-      yield '**';
+      yield n.attributes.marker ?? '**';
       break;
     }
     case 'em': {
-      yield '_';
+      yield n.attributes.marker ?? '*';
       yield* formatInline(formatChildren(n, no));
-      yield '_';
+      yield n.attributes.marker ?? '*';
       break;
     }
     case 'code': {
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -167,11 +167,17 @@ export const thead: Schema = {
 export const strong: Schema = {
   render: 'strong',
   children: ['em', 's', 'link', 'code', 'text', 'tag'],
+  attributes: {
+    marker: { type: String, render: false },
+  },
 };
 
 export const em: Schema = {
   render: 'em',
   children: ['strong', 's', 'link', 'code', 'text', 'tag'],
+  attributes: {
+    marker: { type: String, render: false },
+  },
 };
 
 export const s: Schema = {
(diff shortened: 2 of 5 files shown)


src/formatter.ts
```diff
@@ -305,15 +305,15 @@ function* formatNode(n: Node, o: Options = {}) {
...
-      yield '**';
+      yield n.attributes.marker ?? '**';
       yield* formatInline(formatChildren(n, no));
-      yield '**';
+      yield n.attributes.marker ?? '**';
       break;
     }
     case 'em': {
-      yield '_';
+      yield n.attributes.marker ?? '*';
```
mfix-stripe, line 314: Why are we changing `_` to `*`?

src/schema.ts
```diff
@@ -167,11 +167,17 @@ export const thead: Schema = {
...
 
 export const em: Schema = {
   render: 'em',
   children: ['strong', 's', 'link', 'code', 'text', 'tag'],
+  attributes: {
+    marker: { type: String, render: false },
+  },
 };
 
 export const s: Schema = {
```
mfix-stripe, line 183: Should we add it for Strikethrough (`s`) too?

src/formatter.ts
mfix-stripe: Is there any harm in leaving it as `_` (as a default _convention_)?

mfix-stripe (commented):
PTAL @rpaul-stripe

src/formatter.ts
rpaul-stripe: Yeah, so the issue is that `_` is ignored when it's used inside of a text node, e.g. `foo_bar_baz` is  rendered literally whereas `foo*bar*baz` treats the `*bar*` part as emphasized. Because the underscore behaves differently in this specific case, it's not safe to convert `foo*bar*baz` to `foo_bar_baz`—it changes the meaning of the content. So using the asterisk is a safer universal default than the underscore. It's a little unfortunate, I'd prefer the underscore otherwise.

mfix-stripe: Ahhh, makes sense.

mfix-stripe (approved):
I wonder if we can remove some of the character escape code I wrote now that we have captured the marker. Can you take a look at that briefly @rpaul-stripe?

Issue #371, opened 2023-05-09
Start numbers of ordered lists are ignored


emmatown:
### What happened?

From https://spec.commonmark.org/0.30/#start-number:

> The [start number](https://spec.commonmark.org/0.30/#start-number) of an [ordered list](https://spec.commonmark.org/0.30/#ordered-list) is determined by the list number of its initial list item.

Markdoc currently ignores the start number. I'm assuming this isn't intentional since it isn't mentioned in https://markdoc.dev/docs/faq#is-markdoc-compliant-with-the-commonmark-specification.

### To reproduce

For a Markdoc document like this:

```md
2. something
```

I would have expected the AST to include the start number and the default rendered html to look like this:

```html
<article>
  <ol start="2">
    <li>something</li>
  </ol>
</article>
```

Currently it looks like this:

```html
<article>
  <ol>
    <li>something</li>
  </ol>
</article>
```

### Version

0.2.2

rpaul-stripe:
Thanks for raising this!

It looks like markdown-it captures the number of each list item in the `info` property (apparently [added in 2021](https://github.com/markdown-it/markdown-it/commit/8bcc82aa74164a5e13a104f433c26671a92ed872)), so I can update the parser to expose this to the AST.

For the default schema, I can make it so that the ordered `list` node walks the tree looking for the first list item and then uses its number for the `start` attribute. I'll probably also need to look at the formatter to see if we can incorporate it there.

PR #374, opened 2023-05-10
Store `start` on ordered lists rather than `value` on ordered list items

emmatown:
The way that #371 was fixed in #372 somewhat surprised me because markdown-it returns the start as an attribute on the ordered list so it can be an attribute directly on the list rather than as a new attribute on list items and adding the attribute to the list in a transform. This PR implements that approach of `start` as an attribute on ordered lists rather than `value` on ordered list items.

I've also changed the formatter to not use the number on list items so it only uses the start for the first item and `1` for all other items. For example, for this Markdoc:

```md
2. a
7. b
20. c
```

The formatter in this PR outputs this:

```md
2. a
1. b
1. c
```

Not preserving the numbers of the non-first items aligns with what other formatters do. For example, for the input above, Prettier outputs this:

```md
2. a
3. b
4. c
```

but for this:

```md
2. a
1. b
4. c
```

Prettier outputs:

```md
2. a
1. b
1. c
```

Prettier's rule is "incrementing unless the number in the second list item is 1 then use 1 for everything except for the first which uses the start." Personally, I liked how Markdoc's formatter before #372 always outputted the diff-friendly style of always `1` so I've kept it like that rather than sometimes doing incrementing.

### DIFF
--- a/spec/marktest/tests.yaml
+++ b/spec/marktest/tests.yaml
@@ -15,8 +15,23 @@
     3. baz
   expected:
     - tag: ol
+      children:
+        - tag: li
+          children: [foo]
+        - tag: li
+          children: [bar]
+        - tag: li
+          children: [baz]
+
+- name: Ordered list with start
+  code: |
+    2. foo
+    3. bar
+    4. baz
+  expected:
+    - tag: ol
       attributes:
-        start: '1'
+        start: 2
       children:
         - tag: li
           children: [foo]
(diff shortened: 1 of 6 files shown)


segphault (approved):

rpaul-stripe (approved):

rpaul-stripe:
This looks good, I missed that the list open token had a start attribute. We may at some point in the future still want to capture the number used for each individual item and expose an option to have the formatter output it the way it is expressed in the source, but always using `1` is a fine default.

Issue #379, opened 2023-05-12
Feature request: validating attribute value with custom `matches` function


mskri:
I have been validating attributes with [zod](https://zod.dev/) in the `validate` function.  It would be awesome to be able to define the validation in the attribute definition itself. One option would be to add `validate` function or make the `matches` function return the value of the attribute. You could then use the function to define custom validation in the same place as the attribute itself is declared. It would make it easier to understand what the attribute is and how it should be validated when the declaration and validation are together. 

Yes, you can use RegExp or a list of values to validate the attribute but in some cases that's not enough and it would be better if you could use e.g. zod to do the validation. In some cases the RexExp validation returns error message that isn't useful if you don't understand RexExp. For example if I'm validating that string attribute is not blank using RegExp.

```ts
src: {
  type: String,
  required: true,
  matches: /.+/,
}
```

Resulting error message is `Attribute 'src' must match /.+/. Got '' instead.`. You would have to understand that in RexExp `.+` means "1 or more characters".

**Here's an example of how I would currently do attribute validation with zod**

```ts
export const emailField = {
  render: 'Email',
  children: [],
  attributes: {
    type: {
      type: String,
      required: true,
    },
  },
  validate(node) {
    const errors: ValidationError[] = [];
  
    const email = emailSchema.safeParse(node.attributes.email);
    if (!email.success) {
      errors.push({
        id: 'attribute-value-invalid',
        level: 'error',
        message: "Attribute 'email' has invalid email format",
      });
    }
    
    return errors
  }
};
```

**Here's how I would love to write the validation**

```ts
export const emailField = {
  render: 'Email',
  children: [],
  attributes: {
    email: {
      type: String,
      required: true,
      validation: (value) => {
        const email = emailSchema.safeParse(value);
        if (!email.success) {
          return {
            id: 'image-empty-src',
            level: 'error',
            message: "Attribute 'src' should not be empty.",
          } satisfies ValidationError;
        }
        
        return true;
      },
      errorLevel: 'error'
    },
  }
};
```

PS. What's the purpose of passing `matches(config)`? It includes the whole config and not just the one for the tag.

rpaul-stripe:
I'm open to adding support for this, I don't think there's any real downside in making it possible to define custom validation functions directly on the attribute. One of the reasons we haven't added this yet, however, is that it's possible to achieve this with custom validation types, which are a little bit more powerful.

[...]

PR #380, opened 2023-05-12
Adds validate function to attributes

rpaul-stripe:
This PR addresses issue #379. It adds support for writing a custom `validate` function for an individual schema attribute. The function can return an arbitrary number of Markdoc validation error objects.

After weighing the tradeoffs, I decided not to support async functions for this case because it would require us to collect and await all of them before continuing, introducing more complexity. We can still potentially add support for that in the future if there's sufficient demand.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -115,6 +115,7 @@ export type SchemaAttribute = {
   default?: any;
   required?: boolean;
   matches?: SchemaMatches | ((config: Config) => SchemaMatches);
+  validate?(value: any, config: Config): ValidationError[];
   errorLevel?: ValidationError['level'];
 };
 
--- a/src/validator.test.ts
+++ b/src/validator.test.ts
@@ -289,6 +289,44 @@ bar
   });
 
   describe('attribute validation', () => {
+    describe('with a validate function', () => {
+      it('using simple conditional', () => {
+        const schema = {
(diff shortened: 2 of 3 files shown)


matv-stripe (approved):

mfix-stripe (commented):
@rpaul-stripe I thought we chatted about this previously. I feel like this unnecessarily expands our API surface area, when custom `type`s already support this case, and lead to better reuse. 

@matv-stripe @nkohari-stripe what do you think about this proposal?

rpaul-stripe:
@mfix-stripe I think there's value in being able to do one-off validation functions independently of the type attribute without having to create a reusable type. And I think it's consistent with the way that the validation function works on the schema object, so it doesn't feel like it clutters up the API. We can provide guidance to encourage people to still use custom types in cases where reuse is desirable.

I have been thinking a lot about how we want to handle complex object/array attribute values. Rather than adding a bunch of complexity to Markdoc's type system to explicitly support more feature-rich validation on nested values, what I want to do is provide a pattern that makes it really easy for users to pull in Zod or JSON Schema for those cases. I think a `validate` function on the attribute is ultimately a good way to handle this.

mfix-stripe (approved):
Ack, won't block your vision then 👍

PR #388, opened 2023-05-24
Fix loose lists

### LINKED ISSUE #382: Formatter outputs loose lists where every item only has a single paragraphs as tight lists
### What happened?

From https://spec.commonmark.org/0.30/#loose:

> A list is [loose](https://spec.commonmark.org/0.30/#loose) if any of its constituent list items are separated by blank lines, or if any of its constituent list items directly contain two block-level elements with a blank line between them. Otherwise a list is [tight](https://spec.commonmark.org/0.30/#tight). (The difference in HTML output is that paragraphs in a loose list are wrapped in \<p> tags, while paragraphs in a tight list are not.)

The formatter turns loose lists where every item only has a single paragraphs into tight lists. This also has an larger semantic impact in Markdoc than some extra `<p>` tags since it changes attributes from being on a paragraph to being on a list item.

### To reproduce

For this input:

```md
- a

- b
```

Which turns into this HTML:

```html
<article>
  <ul>
    <li><p>a</p></li>
    <li><p>b</p></li>
  </ul>
</article>
```

The formatter outputs this:

```md
- a
- b
```

Which has this HTML output:
```html
<article>
  <ul>
    <li>a</li>
    <li>b</li>
  </ul>
</article>
```

### Version

0.2.2

### Additional context

_No response_

mfix-stripe:
@rpaul-stripe this fixed issue #382, but does change the effects of the formatter to keep loose lists, loose. LMK what you think.

Closes #382

### DIFF
--- a/src/formatter.test.ts
+++ b/src/formatter.test.ts
@@ -506,6 +506,12 @@ Yes!
 
   it('"loose" lists', () => {
     const source = `
+- a
+
+- b
+
+---
+
 - One
 
   My first paragraph
@@ -528,6 +534,12 @@ Yes!
 
   > Indented blockquote`;
     const expected = `
+- a
+
+- b
+
+---
+
 - One
 
   My first paragraph
@@ -540,6 +552,7 @@ Yes!
   \`\`\`
   {% $code %}
   \`\`\`
+
 - Two
 
   My second paragraph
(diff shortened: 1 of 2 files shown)


rpaul-stripe (approved):
This looks fine to me. 👍

PR #455, opened 2023-10-12
Adds support for dropping location information in `parse` function

rpaul-stripe:
This PR introduces a new parse option called `location` — when set to `false`, it causes the parser to not add the `location` property to document `Node` instances and it leaves the `lines` property as an empty array. This is useful in cases where you are caching the parsed AST in memory for use in production and do not need the location information or want the memory overhead that it incurs.

The PR also updates the `Config` type to add a `validation.environment` property. This property can be populated by the end user in order to specify where validation is occurring, so that they can make their custom validation functions perform different behavior in e.g. CI vs the Markdoc language server.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@markdoc/markdoc",
   "author": "Ryan Paul",
-  "version": "0.3.2",
+  "version": "0.3.3",
   "description": "A text markup language for documentation",
   "main": "dist/index.js",
   "module": "dist/index.mjs",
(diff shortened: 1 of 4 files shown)


mfix-stripe:
@rpaul-stripe what do you think about other non-boolean APIs for handling this, e.g.:
```js
omit: ['location']
```
or 
```js
omit: {
  location: true
}
```
If it is truly about eliminating keys from an object, this feels more future proof.

rpaul-stripe:
> @rpaul-stripe what do you think about other non-boolean APIs for handling this, e.g.:

We are quite limited in what keys we can remove from `Node` without breaking backwards compatibility. I think it is unlikely that we will extend this to more keys in the foreseeable future. I prefer just "location" for now.

mfix-stripe (approved):

PR #463, opened 2023-11-08
Fix table cell widths in formatter

nkohari-stripe:
This patch changes the behavior of `Markdoc.format()` for tables. Instead of setting the width of each table cell to the maximum width of all cells in the _table_, this patch will set the width of each table cell to the maximum width of all cells in the _column_. This makes tables significantly more condensed and easier to read.

(It also indirectly fixes a scenario where a `TypeError` was being thrown in cases where `Array.reduce()` was being called for an empty array and with no initial value.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@markdoc/markdoc",
-  "version": "0.3.4",
+  "version": "0.3.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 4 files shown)


mfix-stripe (approved):
LGTM! Thanks for the update @nkohari-stripe!

mfix-stripe:
@nkohari-stripe before you merge, want to bump the patch version in `package.json` and `package-lock.json` and then we can cut a new release?

PR #472, opened 2023-11-28
Changes th width property to string type in schema

### LINKED ISSUE #471: Table's width attribute throws type error on strings
### What happened?

When working with Table tags, the `width` attribute throws a type error if the value is not a Number. However, the `width` attribute works (and is expected to work) with strings. In the [Markdoc Attributes documentation](https://markdoc.dev/docs/attributes), there's an example of using the `width` attribute inside a Table with a percentage value, which is a String. The syntax works, but throws an error. 

To clarify:
- Works as expected → a Table's `width` attribute accepts a String.
- Unexpected behavior → passing a String (i.e. `25%`) in a Table's `width` attribute throws the following error:
```
Attribute 'width' must be type of 'Number'
```

### To reproduce

1. Go to the official [Markdoc try it sandbox](https://markdoc.dev/sandbox)
2. Paste the following table in the bottom of the editor:
```
{% table %}
* Heading 1 {% width="20%" %}
* Heading 2
---
* Row 1 Cell 1
* Row 1 Cell 2
{% /table %}
```
3. Hover over the Heading 1 line to see error

(Note that despite the error, the `20%` syntax _does work_ and limits the table's column to 20%.)

### Version

_No response_

### Additional context

Screenshot 2023-11-28 at 11 30 07 AM

rpaul-stripe:
The schema for the `th` node was written to expect a number for the `width` attribute, but when the `width` attribute is rendered literally in markup it needs to be a string. This PR changes the schema and also increments the package's version so we can push an update.

Closes #471

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@markdoc/markdoc",
   "author": "Ryan Paul",
-  "version": "0.3.6",
+  "version": "0.4.0",
   "description": "A text markup language for documentation",
   "main": "dist/index.js",
   "module": "dist/index.mjs",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@markdoc/markdoc",
   "author": "Ryan Paul",
-  "version": "0.3.5",
+  "version": "0.3.6",
   "description": "A text markup language for documentation",
   "main": "dist/index.js",
   "module": "dist/index.mjs",
--- a/spec/marktest/tests.yaml
+++ b/spec/marktest/tests.yaml
@@ -712,7 +712,7 @@
 - name: Advanced table format
   code: |
     {% table %}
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@markdoc/markdoc",
   "author": "Ryan Paul",
-  "version": "0.3.5",
+  "version": "0.3.6",
```
mfix-stripe, line 4: We should probably release this as a new minor version right? Just in case it breaks folks' validation.

mfix-stripe (commented):
PTAL  @rpaul-stripe

rpaul-stripe:
@mfix-stripe changed it from 0.3.6 to 0.4.0.

mfix-stripe (approved):

Brandylee24 (approved):

PR #473, opened 2023-11-30
Create devcontainer.json

Gmrhom:


rpaul-stripe:
Markdoc has a very simple dependency stack and doesn't require any specialized configuration to run in a container. I don't think there's much real value in adding this and I would prefer not to add a lot of unnecessary configuration files in the repository root.

PR #474, opened 2023-12-06
Fix escaping image src in formatter

emmatown:
The regex here was missing the global flag (The one used for links already has the global flag)


rpaul-stripe (approved):
Thanks!

Issue #488, opened 2024-03-06
Replace the Tag class by a POJO to make the RenderableTreeNode serializable 


CapitaineToinon:
### What happened?

I understand my use case is pretty specific and there are workarounds but that would be lovely it this could be fixed.

I'm trying to use Markdoc in an SSR app where the Markdoc.parse and Markdoc.transform are done on the server. The renderable node is then sent to the client and rendered there. With Remix, everything's fine since they use JSON.stringify and JSON.parse under the hood.

However if you try to use Markdoc with sveltekit, you won't be able to do that as they use devalue to get the data around. It can be argued that sveltekit should add an option to either add custom reducers for devalue or an option to opt out but at the same time, looking at the implementation of the Tag class, it seems to be a simple wrapper around some values and could easily be replaced by a POJO. I'd be up to create a PR for that.

What to you guys thing?

### To reproduce

Just forked the devalue svelte repl and included some Markdoc example:

https://svelte.dev/repl/f9f7c58c7a9d440f84f11ef697ee8bdb?version=3.49.0

### Version

0.4.0

rpaul-stripe:
Hey @CapitaineToinon,

Using the class is actually completely optional. You can write your `transform` functions in your project so that they return a POJO with the shape `{$$mdtype: 'Tag', name: 'foo', attributes: {}, children: []}` and it should work exactly the same way. The `Tag.isTag` function checks the `$$mdtype` property rather than looking to see if a value is a tag instance, so it will still work as expected.

I'm not really familiar with svelte or devalue, but I would assume you would be able to manually serialize the renderable tree to JSON somewhere before it hits the framework. You can just wrap `JSON.stringify` around the output of `Markdoc.transform` and then call `JSON.parse` on it before rendering.

At Stripe, we also serialize our renderable tree to JSON on the server side and render it on the client — Markdoc was actually designed specifically to support that use case. I'm not sure I understand why that wouldn't work with Svelte, maybe there's something I'm missing about how it does SSR.

In any event, I don't think we can accept the PR as written because eliminating the `Tag` class would be a significant non-backwards-compatible change that would require everyone to rewrite their custom tag definitions.

CapitaineToinon:
[...]

PR #489, opened 2024-03-06
Migrating Tag away from classes

CapitaineToinon:


CapitaineToinon:
Not sure about the syntax. Also I guess this would be a breaking change, although would could keep the class constructor and just convert to a POJO in the transform instead. Not gonna bother to do more for now, this is kinda just a proof of concept. 

I installed my markdoc version from this PR on my sveltekit app mentionned in #488 and it fixes the issues flawlessly. Now the `RenderableTreeNode` is ensured to be a POJO and can be safely serialized.

rpaul-stripe:
Closing, per discussion here: https://github.com/markdoc/markdoc/discussions/490?converting=1#discussioncomment-8720924

PR #492, opened 2024-03-25
Fix Typescript error for ^5.4

dcousens:
As seen in https://github.com/keystonejs/keystone/pull/9060

```typescript
error TS2866: Import 'SyntaxError' conflicts with global value used in this file, so must be declared with a type-only import when 'isolatedModules' is enabled.

1 import { parse, SyntaxError } from './grammar/tag';
```


dcousens:
Wait, `SyntaxError` is used later with `instanceof SyntaxError`, sorry

YevhenSoldatov1998:
I have the same problem. How fix it ?

PR #506, opened 2024-05-15
Fix inconsistency between type libraries for markdown-it and linkify-it

nkohari:
Hey there! 😄 

I was updating [apocrypha](https://github.com/nkohari/apocrypha) to the latest version of Markdoc and started running into some type errors:

```
> @apocrypha/core@0.3.0 build:types
> tsc --emitDeclarationOnly --outDir dist

../../node_modules/@types/markdown-it/lib/index.d.ts:151:33 - error TS2694: Namespace '"/Users/nkohari/Work/apocrypha/node_modules/@types/linkify-it/index"' has no exported member 'LinkifyIt'.

151     readonly linkify: LinkifyIt.LinkifyIt;
                                    ~~~~~~~~~

Found 1 error in ../../node_modules/@types/markdown-it/lib/index.d.ts:151
```

Turns out, this is because `@types/markdown-it` depends on any version (`*`) of `@types/linkify-it`, but the latter wasn't reverse-compatible. This patch pins the version of `@types/linkify-it` to the one which is compatible with the the version of the actual `markdown-it` library that Markdoc uses.

Let me know if you have any questions or need more info!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@markdoc/markdoc",
-  "version": "0.3.5",
+  "version": "0.4.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@markdoc/markdoc",
-      "version": "0.3.5",
+      "version": "0.4.0",
       "license": "MIT",
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


matv-stripe (approved):
Thanks!

Issue #477, opened 2024-01-05
Update Config types to account for Schema render function


herrethan:
### What happened?

The `render` function for tags and nodes accepts a generic type, [as defined in the Schema](https://github.com/markdoc/markdoc/blob/60a2c831bd7ac8f2f24aabfde0b36e56e5d0dbe1/src/types.ts#L101-L109).

This allows us to pass in custom components to the render function, as [suggested here](https://markdoc.dev/docs/nextjs#tags) when using nextjs. 

For example:
```ts
// tags/callout.markdoc.ts

import { Callout } from '../components/callout';

export const callout: Schema<Config, typeof Callout> = {
  render: Callout,
  children: ['paragraph', 'tag', 'list'],
};
```

However, the `Config` expects the render function to be a string only. This is also documented as a string only [here](https://markdoc.dev/docs/tags#options), so when trying to create a config using the above approach you get:

```
Type '() => Element' is not assignable to type 'string'
```

Is there a better way to type the Config I'm not seeing? Or should the Config types be updated to accommodate custom render components in Schema?

Great library BTW!

### To reproduce

On versions:    
"@markdoc/markdoc": "^0.4.0",
"@markdoc/next.js": "^0.3.6",

### Version

0.4.0

PR #509, opened 2024-05-23
Add type generic to Config

mfix-stripe:
@segphault I'm not sure if we want to convolute our types in this way just to support Next.js (I'm actually leaning towards no…), but wanted to throw up this PR for consideration to solve https://github.com/markdoc/markdoc/issues/477 

If you don't think we should go this direction, can you comment in the [original thread](https://github.com/markdoc/markdoc/issues/477) and close it? 

Closes https://github.com/markdoc/markdoc/issues/477

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -20,10 +20,10 @@ export type AttributeValue = {
   value: any;
 };
 
-export type Config = Readonly<ConfigType>;
+export type Config<R = string> = Readonly<ConfigType<R>>;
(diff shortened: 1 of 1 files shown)


matv-stripe (approved):
I think this is a somewhat common pattern, I would be okay with merging this

segphault:
I think I'm good with this too. Sorry for the delay looking at it!

PR #523, opened 2024-09-25
Add orderedListMode option to Markdoc.format

bibek-stripe:
Adds an `orderedListMode` option to the `Options` accepted by `Markdoc.format` such that when set to `'increment'`, calling `Markdoc.format` on the AST resulting from parsing

```
1. Foo
2. Bar
3. Baz
```

would not be changed to

```
1. Foo
1. Bar
1. Baz
```

and similarly, the latter as input WOULD get formatted to the former.

### DIFF
--- a/src/formatter.test.ts
+++ b/src/formatter.test.ts
@@ -506,6 +506,48 @@ Yes!
     stable(expected);
   });
 
+  it('human-readable ordered lists', () => {
+    const source = `
+- foo
+- bar
+* baz
+* qux
+
+
+7) foo
+1) bar
+1) baz
+3. foo
+1. bar
+1. baz
+1) foo
+4) bar
+9) baz
+`;
+    const expected = `
+- foo
+- bar
+
+* baz
+* qux
+
+7) foo
+8) bar
+9) baz
+
+3. foo
+4. bar
+5. baz
+
+1) foo
+2) bar
+3) baz
+`;
+    const options = {increasingOlMarkers: true};
+    check(source, expected, options);
+    stable(expected, options);
+  });
+
   it('"loose" lists', () => {
     const source = `
 - a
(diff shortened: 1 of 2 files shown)


src/formatter.test.ts
```diff
@@ -506,6 +506,48 @@ Yes!
     stable(expected);
   });
 
+  it('human-readable ordered lists', () => {
```
matv-stripe:
```suggestion
  it('increment list numbers option', () => {
```

matv-stripe (approved):

src/formatter.ts
```diff
@@ -5,6 +5,7 @@ import type { AttributeValue, Function, Node, Value, Variable } from './types';
 type Options = {
   allowIndentation?: boolean;
   maxTagOpeningWidth?: number;
+  incrementListNumbers?: boolean;
```
mfix-stripe, line 8: Lurk: I wonder if this parameter should be a string union type instead? Something like: `orderedListMode: "increment"` or something?

mfix-stripe: Cc @bibek-stripe @matv-stripe

matv-stripe: I like that too!

matv-stripe: @mfix-stripe do you have other modes in mind?

src/formatter.test.ts
bibek-stripe: updated it in a commit while you were typing!

src/formatter.ts
mfix-stripe: Not at the moment, I just always tend to lean away from booleans for future-proofing. I could see someone wanting something else in the future.

bibek-stripe:
Updated to use a string union! ptal @mfix

mfix-stripe (approved):
🚀

PR #524, opened 2024-09-26
Add parent getter

mfix-stripe:
Cc @segphault 
Cc @bibek-stripe

### DIFF
--- a/src/ast/node.test.ts
+++ b/src/ast/node.test.ts
@@ -74,6 +74,7 @@ baz
       'inline',
       'text',
     ]);
+    expect(iter[1].parent).toEqual(iter[0]);
   });
 });
 
--- a/src/ast/node.ts
+++ b/src/ast/node.ts
@@ -42,8 +42,19 @@ export default class Node implements AstType {
     this.slots = {};
   }
 
+  get parent(): Node | null {
+    return null;
+  }
+
   *walk(): Generator<Node, void, unknown> {
-    for (const child of [...Object.values(this.slots), ...this.children]) {
+    const parent = this;
+    for (let child of [...Object.values(this.slots), ...this.children]) {
+      Object.defineProperty(child, 'parent', {
+        configurable: true,
+        get() {
+          return parent;
+        },
+      });
       yield child;
       yield* child.walk();
     }
--- a/src/ast/node.ts
+++ b/src/ast/node.ts
@@ -47,8 +47,9 @@ export default class Node implements AstType {
   }
 
   *walk(): Generator<Node, void, unknown> {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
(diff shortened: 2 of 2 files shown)


segphault:
I'm leaning towards no on this change, but I'd like to hear more about the problem that you're looking to solve with it before I make a decision. 

We generally discourage users from implementing behaviors in transforms that rely on a node's parentage. This makes it easier to operate on subsets of the tree and avoids a bunch of problems that can arise from copying and moving nodes around during transform steps. There are, of course, a bunch of valid reasons why you might want to do this for static analysis outside of a transform.

Are you aware of the [`walkWithParents`](https://github.com/markdoc/markdoc/blob/7bbb0b39e829fd028f1a295d2a33cfd38bc5015e/src/validator.ts#L297-L304) function that is used for this in validation? I'm wondering if it is sufficient for what you are trying to do here. I'm open to maybe exposing that as a top-level export.

mfix-stripe:
> Are you aware of the [`walkWithParents`](https://github.com/markdoc/markdoc/blob/7bbb0b39e829fd028f1a295d2a33cfd38bc5015e/src/validator.ts#L297-L304) function that is used for this in validation? I'm wondering if it is sufficient for what you are trying to do here. I'm open to maybe exposing that as a top-level export.

Was not aware of this! @bibek-stripe were you? This would solve your use case right?

bibek-stripe:
My use case is that I'd like to validate that a `card` tag is only rendered as a child of `card-group`. In other words, a document with contents

```
{% card-group %}
{% card %}Foo{% /card %}
{% /card-group %}
```

is valid but

```
{% card %}Foo{% /card %}
```

should surface an error.

I'm currently doing this validation all the way up at the `document` node and walking the entirety of the document AST to check for all instances of this which is actually working quite well for my use case.

I wasn't aware of the `walkWithParents` method. It looks like it's only importable from `@markdoc/markdoc/dist/src/validator` so probably not safe to import right? Also, how would one use it for a use case like mine?

segphault:
Ah! If you are using this only in validation, that is already supported. In a validation function, the property `config.validation.parents` has an array with all of the parents of the current node. I wrote a [blog post](https://seg.phault.net/blog/2024/04/accessing-parent-nodes-in-a-markdoc-validation-function/) that demonstrates how to use it with a similar use case to the one you describe. 

You want to do something like this in your validation function: `config.validation.parents.some(parent => parent.tag === 'card-group')`

bibek-stripe:
Oh sweet! Would be cool to have this in the docs at markdoc.dev as well! :)

mfix-stripe:
> Oh sweet! Would be cool to have this in the docs at markdoc.dev as well! :)

@bibek-stripe feel free to add it and I can review

PR #537, opened 2025-01-15
Fix formatting fences when content has no final newline

jstasiak:
Formatting a fence node with content set to "foo" previously produced

    ```
    foo```

which doesn't parse correctly (the backticks are included in the content; if there is more markup following such fence block the markup can also be misinterpreted as the contents).

### DIFF
--- a/src/formatter.test.ts
+++ b/src/formatter.test.ts
@@ -843,4 +843,9 @@ ${'`'.repeat(4)}
     const d = diff(expected, b.trim());
     if (d && d.includes('Compared values have no visual difference.')) return;
   });
+
+  it('makes sure fences are formatted correctly if content has no ending newline', () => {
+    const node = new Markdoc.Ast.Node('fence', {content: 'foo'})
+    expect(format(node)).toEqual('```\nfoo\n```\n')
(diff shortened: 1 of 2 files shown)


jstasiak:
I bumped into this problem while working on auto-generating Markdoc content programmatically.

Can content without the final newline be considered a programming error? I assumed no, hence the PR.

Let me know if I need to adjust the code style in any way.

matv-stripe (approved):

matv-stripe:
@jstasiak , when we programmatically create Markdoc elements, we usually just add the newline into content ourselves when we create fence nodes. This is not intuitive though, so I am +1 to your change.

jstasiak:
No problem, happy to contribute – that's an impressively quick review BTW :D

jstasiak:
> when we programmatically create Markdoc elements, we usually just add the newline into content ourselves when we create fence nodes

Yeah, fair – I realized I had to do that in the end, it just took me longer than I'm willing to admit to figure out why the fences were "eating" some of the content that followed them and what the solution was.

Either way happy to see this accepted – thank you.

Issue #539, opened 2025-01-17
type-only exports make some code somewhat inconvenient


jstasiak:
Hey, not sure if this should be a discussion or an issue – please convert if the former.

The problem that I bumped into is I can't have code like the following

```typescript
import { Node } from '@markdoc/markdoc'

function f(): Node {
    // error below: 'Node' cannot be used as a value because it was exported using 'export type'.ts(1362)
    return new Node('tag', {}, [], 'foo')
}
```

because `Node` is `export type`-exported:

```
// markdoc/index.ts
export * from './src/types';

// markdoc/src/types.ts
import type Func from './ast/function';
import type Node from './ast/node';
import type Var from './ast/variable';
import type Tag from './tag';

export type { Node, Tag };
export declare type Function = Func;
export declare type Variable = Var;
```

Instead I have to do this, which I'm not fully happy with because a) verbosity b) mismatch between the return type declaration and the type instantiated:

```typescript
import Markdoc, { Node } from '@markdoc/markdoc'

function f(): Node {
    return new Markdoc.Ast.Node('tag', {}, [], 'foo')
}
```

This is with @markdoc/markdoc 0.4.0.

Would you be willing to accept a PR that changes the exports to allow runtime of `Node` and maybe a few other classes? Or is instantiating via `Markdoc.Ast.Node` etc. definitely the way to go here?

mfix-stripe:
I run into this issue and I personally would be in favor of exporting the class directly. @segphault you against this?

segphault:
I'm open to accepting a PR that addresses this. 👍

jstasiak:
#540 addresses this – since I only ever had a problem with `Node` specifically I limited the scope to just `Node`; if there are other classes that you think should get the same treatment I can adjust.

PR #540, opened 2025-01-20
Re-export Node in a way that allows runtime use

jstasiak:
Previously it wasn't possible to have code like

    import { Node } from '@markdoc/markdoc'

    function f(): Node {
        // error below: 'Node' cannot be used as a value because it was exported using 'export type'.ts(1362)
        return new Node('tag', {}, [], 'foo')
    }

This patch rectifies that and we can have the same type used in return type position and the class instantiation position.

Resolves: https://github.com/markdoc/markdoc/issues/539

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,4 +1,5 @@
 import Ast from './src/ast';
+import Node from './src/ast/node';
 import format from './src/formatter';
 import functions from './src/functions';
 import parser from './src/parser';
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -113,6 +114,7 @@ export {
   transforms,
   renderers,
   Ast,
+  Node,
```
jstasiak, line 117: I was afraid

```
export * from './src/types';
```

was going to clash with this (re)export but after some preliminary tests I think it's fine.

segphault (approved):

Issue #573, opened 2025-07-21
Support Markdoc auto-links


rparman-stripe:
### What happened?

Normal _GitHub-Flavored Markdown_ tables are formatted like this:

```markdown
| Tool | Endpoint |
|-|-|
| Homepage  | <https://markdoc.dev> |
| GitHub  | <https://github.com/markdoc/markdoc> |
```

The canonically-Markdown way to list URLs is to wrap them in `<>` (like the example above).

Unfortunately, when I leverage Markdoc’s auto-format functionality, it converts the URLs into Markdown _links_, which is _duplicative_ and _unnecessary_.

(Note that my usage of the terms _URLs_ and _links_ are different.)

```markdown
| Tool     | Endpoint                                                                 |
|----------|--------------------------------------------------------------------------|
| Homepage | [https://markdoc.dev](https://markdoc.dev)                               |
| GitHub   | [https://github.com/markdoc/markdoc](https://github.com/markdoc/markdoc) |
```

I end up _manually_ reverting to the canonical Markdown format for URLs.

```markdown
| Tool     | Endpoint                             |
|----------|--------------------------------------|
| Homepage | <https://markdoc.dev>                |
| GitHub   | <https://github.com/markdoc/markdoc> |
```

### To reproduce

1. Create a new page in the internal Stripe documentation tool which supports Markdoc syntax.
2. Create a table like in the first example above.
3. Click the auto-format button.

### Version

0.3.4

### Additional context

* [Original Markdown spec](https://daringfireball.net/projects/markdown/syntax), under _Miscellaneous_ → _Automatic Links_
* [Commonmark spec](https://spec.commonmark.org/0.31.2/#autolinks), under _6.5 Autolinks_
* [GitHub-Flavored Markdown spec](https://github.github.com/gfm/#autolinks), under _6.8 Autolinks_

matv-stripe:
To support this, we should be able to check that  `n.children[0].attributes.content === n.attributes.href` and if so output `<` n.attributes.href`>`

https://github.com/markdoc/markdoc/blob/main/src/formatter.ts#L193-L204

PR #574, opened 2025-07-21
Support auto-links

mfix-stripe:
Closes https://github.com/markdoc/markdoc/issues/573 

PTAL  @matv-stripe WDTY?


matv-stripe (approved):
nice

PR #609, opened 2026-03-14
Fix: escaped links/images/autolinks lose backslash in Markdoc.format (fixes #590)

### LINKED ISSUE #590: Escaped links/images lose escaping after formatting
### What happened?

Formatting markdoc like `\[a](https://example.com)` formats to `[a](https://example.com)`. It removes the escaping when it should preserve it.

### To reproduce

For all of these examples, when rendered, they correctly render the escaped content

```md
\[a](https://example.com)

\!\[a](https://example.com)

\<https://example.com>
```

When passed through `Markdoc.format` though, the result is

```md
[a](https://example.com)

![a](https://example.com)

<https://example.com>
```

with the escaping removed which shouldn't happen

### Version

0.5.4

### Additional context

_No response_

shivamtiwari3:
## Summary

Fixes #590.

`Markdoc.format` was emitting text nodes verbatim, so a text node whose content happened to be `[a](url)` (produced by parsing `\[a](url)`) was written back as `[a](url)` — which the parser re-reads as a link. The original escaping was silently dropped on every format round-trip.

---

## Root Cause

In `formatter.ts`, the `'text'` case of `formatNode` (previously line 214) passed `content` directly to `escapeMarkdownCharacters`. That helper only re-escaped characters that could form headings, blockquotes, or emphasis — it had no knowledge of inline-link or autolink syntax.

Three patterns broke:

| Input | Stored text-node content | Old output | Re-parsed as |
|-------|--------------------------|------------|--------------|
| `\[a](url)` | `[a](url)` | `[a](url)` | **link** ✗ |
| `\!\[a](url)` | `a` | `a` | **image** ✗ |
| `\<https://example.com>` | `<https://example.com>` | `<https://example.com>` | **autolink** ✗ |

---

## Solution

Three targeted guards are applied to `content` **before** the existing per-character escaping:

1. **`startsWithLink(content, offset)`** — a small scanner that follows CommonMark's balanced-parenthesis rule for link destinations. It escapes `[` only when it genuinely opens an inline link `[text](url)`. Patterns such as `Image` — which the parser already treats as plain text — are left untouched.

2. **`/<scheme:\/\/...>/` regex** — escapes `<` that would be re-parsed as a scheme-based autolink (e.g. `<https://example.com>`).

3. **`nextSibling` tracking** — `formatChildren` now forwards the next sibling into each child's options. When a `text` node ends with `!` and its next sibling is a `link` node (the AST shape produced by `\link`), the trailing `!` is escaped to `\!`, preventing the sequence from being re-read as an image.

---

## Testing

- Added `'preserves escaping on links, images, and autolinks (issue #590)'` in `src/formatter.test.ts`:
  - `\[a](url)` round-trips stably (formatter re-emits `\[`)
  - `\<https://example.com>` round-trips stably (formatter re-emits `\<`)
  - `[brackets]` (no URL) gains no spurious backslash
  - `<not-an-autolink>` gains no spurious backslash
  - `\!\[a](url)` re-parses to the same text-node content (`a`)
- All **260 existing specs** still pass.
- Run with: `npm test`

---

## Checklist

- [x] Fixes the root cause (not just the symptom)
- [x] New tests cover the exact scenarios from issue #590
- [x] All existing tests pass (260 specs, 0 failures)
- [x] No unrelated changes
- [x] Code style matches project conventions
- [x] Lint clean (`npm run lint` — no errors)

### DIFF
--- a/src/formatter.test.ts
+++ b/src/formatter.test.ts
@@ -855,4 +855,29 @@ ${'`'.repeat(4)}
     const node = new Markdoc.Ast.Node('fence', { content: 'foo' });
     expect(format(node)).toEqual('```\nfoo\n```\n');
   });
+
+  it('preserves escaping on links, images, and autolinks (issue #590)', () => {
+    // \[text](url) — escaped [ must be re-emitted so it stays plain text.
+    // stable() requires a trailing \n because format appends one for paragraphs.
+    stable(String.raw`\[a](https://example.com)` + '\n');
+
+    // \<url> — escaped autolink must stay plain text, not become an autolink
+    stable(String.raw`\<https://example.com>` + '\n');
+
+    // [brackets] with no following (url) must not gain a spurious backslash
+    stable('Item with [brackets]\n');
+
+    // <not-an-autolink> must not gain a spurious backslash
+    stable('<not-an-autolink>\n');
+
+    // \!\[text](url) — the parser collapses \! and \[ into the single text
+    // node "![a](url)". The formatter escapes the [ (producing !\[...]) which
+    // is semantically equivalent and round-trips to the same content.
+    const ast = Markdoc.parse(String.raw`\!\[a](https://example.com)`);
+    const reparsed = Markdoc.parse(format(ast));
+    const content =
+      reparsed.children?.[0]?.children?.[0]?.children?.[0]?.attributes
+        ?.content;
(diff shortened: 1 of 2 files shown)


shivamtiwari3:
Hi — happy to make any further changes if helpful. Just checking if this is still under consideration.

PR #624, opened 2026-03-31
Disable markdoc tag in link href in tokenizer

rayc-stripe:
Markdoc doesn't have any validation to fail the use of markdoc variables or tags (i.e. `{% %}` format) in link href. 

In this PR, we capture the link format in the tokenizer, and push an error when the tag format is found in the href. 

Another option is to run the validation in [link schema](https://github.com/markdoc/markdoc/blob/main/src/schema.ts#L207-L214), but the problem is that the default markdown parsing doesn't parse the content as link when space is included in the href. For example, `[Link](https://{%tag%})` can be parsed, but `[Link](https://{% tag %})` can't be parsed.

This is why I choose to add a plugin for the tokenizer and throw an error there. This can ensure the default parsing is not changed while additional validation can be enforced.

This link validation is optional. It can be enabled through `allowLinkValidation` in the config. Custom protocol to validate (`validatedProtocols`) can be specified in `linkValidationOptions`. It's default to `http` and `https`.

### DIFF
--- a/src/tokenizer/plugins/link.ts
+++ b/src/tokenizer/plugins/link.ts
@@ -1,0 +1,41 @@
+import type MarkdownIt from 'markdown-it/lib';
+import type StateCore from 'markdown-it/lib/rules_core/state_core';
+import type Token from 'markdown-it/lib/token';
+import type { ValidationError } from '../../types';
+
+import { TAG_PATTERN } from '../../utils';
+
+const MARKDOWN_LINK_DESTINATION = /\[[^\]]+\]\(([^)\n]+)\)/g;
+const INVALID_HREF_MESSAGE =
+  "The 'href' format cannot contain Markdoc tag or variable. URLs must be static strings.";
+
+type TokenWithErrors = Token & { errors?: ValidationError[] };
+
+function pushHrefError(token: Token) {
+  const tokenWithErrors = token as TokenWithErrors;
+  if (!tokenWithErrors.errors) tokenWithErrors.errors = [];
+  tokenWithErrors.errors.push({
+    id: 'href-format-invalid',
+    level: 'error',
+    message: INVALID_HREF_MESSAGE,
+  });
+}
+
+function core(state: StateCore) {
+  let token: Token;
+  for (token of state.tokens) {
+    if (token.type !== 'inline' || typeof token.content !== 'string') continue;
+
+    const matches = token.content.match(MARKDOWN_LINK_DESTINATION) || [];
+    const invalidLink = matches.find((linkSource) =>
+      TAG_PATTERN.test(linkSource)
+    );
+    if (!invalidLink) continue;
+
+    pushHrefError(token);
+  }
+}
+
+export default function plugin(md: MarkdownIt) {
+  md.core.ruler.push('link_url_validation', core);
+}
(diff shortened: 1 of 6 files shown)


src/tokenizer/plugins/link.ts
```diff
@@ -0,0 +1,41 @@
...
+    message: INVALID_HREF_MESSAGE,
+  });
+}
+
+function core(state: StateCore) {
+  let token: Token;
+  for (token of state.tokens) {
+    if (token.type !== 'inline' || typeof token.content !== 'string') continue;
+
+    const matches = token.content.match(MARKDOWN_LINK_DESTINATION) || [];
```
yue-stripe: Using raw token content would incorrectly flag content that isn't rendered as a link e.g.
```
`[Link]({% tag %})`
```
which doesn't render a Link but an `inline code` component with the content.

Additionally the regex won't work with URLs with parentheses like `https://en.wikipedia.org/wiki/Exam_(disambiguation){% tag %}` ([regex](https://regex101.com/?regex=%5C%5B%5B%5E%5C%5D%5D%2B%5C%5D%5C%28%28%5B%5E%29%5Cn%5D%2B%29%5C%29&testString=%5Bwith+parentheses+in+URL%5D%28https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FExam_%28disambiguation%29%7B%25+tag+%25%7D%29&flags=gm&flavor=pcre2&delimiter=%2F))

We should walk the AST produced by `markdown-it` instead of trying to parse markdown links ourselves. Something like

```
function getAttr(token: Token, name: string) {
    return token.attrs?.find(([key]) => key === name)?.[1];
  }

  const href = getAttr(token, 'href');
  if (href && containsMarkdocTag(href)) {
    pushHrefError(token);
  }
```

src/utils.ts
```diff
@@ -12,6 +12,7 @@ enum STATES {
 
 export const OPEN = '{%';
 export const CLOSE = '%}';
+export const TAG_PATTERN = /\{%\s*[\s\S]*?%\}/;
```
yue-stripe: This regex duplicates the markdoc tag scanning logic defined later in this file. Can we make a helper function `containsMarkdocTag` that checks for `OPEN` + `findTagEnd()`, something like

```ts
export function containsMarkdocTag(content: string): boolean {
    for (let pos = 0; pos < content.length; pos++) {
      if (!content.startsWith(OPEN, pos)) continue;

      const end = findTagEnd(content, pos);
      if (end != null) return true;

      // Mirror parseTags(): ignore a dangling `{%` and keep scanning.
      pos += OPEN.length - 1;
    }

    return false;
  }
```

yue-stripe (changes requested):
I would recommend reusing `markdown-it`'s markdown AST instead of trying to parse markdown links ourselves

src/tokenizer/plugins/link.ts
rayc-stripe: `href` is only parsed and stored when the link is valid. I fix the failed test cases by tracing the start position from a the start tag (`%{`) and see if the content contains the `https://` instead.

yue-stripe (approved):

PR #627, opened 2026-06-29
Fix incorrectly escaped number signs in text

### LINKED ISSUE #637: Heading escape regex matches # anywhere in text, not just at line start
### What happened?

Since #583, `#` within block content is incorrectly getting escaped when formatting, e.g. `C# is a language` → `C\# is a language`.

### To reproduce

1. Format a text node whose content contains a `#` followed by whitespace anywhere other than the start of a line:
   ```js
   const { format, Ast } = require('@markdoc/markdoc');
   const node = new Ast.Node('text', { content: 'Text # not a heading' });
   console.log(format(node));
   ```
2. Observe the output is `Text \# not a heading` — the `#` is escaped even though it can't start a heading there.
3. Same happens with `C# is a language` → `C\# is a language`.

As a broken test case: the `escape markdown content` fixture in `src/formatter.test.ts` (~line 148, next to the existing `Text > not a blockquote` line) should include a new line:
```
Text # not a heading
```
This fails against the current regex and passes once `#+\s` is anchored with `^`.

### Version

_No response_

### Additional context

Fix is a one-character change in `src/formatter.ts:227`:
```diff
- yield* escapeMarkdownCharacters(content, /^\*|#+\s|^>/);
+ yield* escapeMarkdownCharacters(content, /^\*|^#+\s|^>/);
```

mahirhir:
Fixes https://github.com/markdoc/markdoc/issues/637

#583 anchored the blockquote `>` escape to the start of a line, since `>` is only a block marker there. The same escape expression still matches the heading `#` anywhere in the text:

```js
escapeMarkdownCharacters(content, /^\*|#+\s|^>/)
```

So a `#` that follows other text gets escaped even though it can't start a heading:

- `Text # not a heading` → `Text \# not a heading`
- `C# is a language` → `C\# is a language`

This anchors `#+\s` to the start of the line as well, matching `^\*` and `^>`. A leading `#` in literal text is still escaped, so `\# Heading` keeps round-tripping unchanged.

The fixture in the existing "escape markdown content" test gains a `Text # not a heading` line, which fails before the change.

### DIFF
--- a/src/formatter.test.ts
+++ b/src/formatter.test.ts
@@ -147,6 +147,8 @@ regular_word_with_underscores
 
 Text > not a blockquote
 
+Text # not a heading
+
 \\# Heading
 
 \\### Heading
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -224,7 +224,7 @@ function* formatNode(n: Node, o: Options = {}) {
           yield* escapeMarkdownCharacters(content, /[*_~]/g);
         } else {
           // Escape > blockquote, * list item, and heading
-          yield* escapeMarkdownCharacters(content, /^\*|#+\s|^>/);
+          yield* escapeMarkdownCharacters(content, /^\*|^#+\s|^>/);
         }
       }
 


joelzwarrington:
:wave: Hi @mahirhir, thank you for the contribution and reporting this bug! I've created [an issue](https://github.com/markdoc/markdoc/issues/637) to report the issue for better visibility. I'd love to merge your changes but we require a signed contributer license agreement. Can you do this please?

mahirhir:
@joelzwarrington thank you for picking this up and for opening #637.

On the CLA, I want to give you a straight answer rather than leave the thread open on a maybe: **signing is not a decision I am in a position to make on my own**, so I cannot give you a date and I would rather say that than have you keep this in your queue waiting on me.

So please do not wait on me. **Take the fix directly.** It is one character, the `^` anchor on the heading alternative:

```diff
-          yield* escapeMarkdownCharacters(content, /^\*|#+\s|^>/);
+          yield* escapeMarkdownCharacters(content, /^\*|^#+\s|^>/);
```

Still at `src/formatter.ts:227` on `main` as of today. The other two alternatives are already anchored; the heading one was the only one that was not, so a `#` anywhere in a line was treated as a heading marker.

Behaviour of the two patterns, so the blast radius is visible (`true` means the text gets escaped):

```
  input                     current  fixed
  "Text # not a heading"    true     false     <- the bug
  "# real heading"          true     true
  "> quote"                 true     true
  "*bold"                   true     true
  "a > b"                   false    false
```

Only the first row changes. Every case that should be escaped still is.

The test in this PR is two lines and can be lifted as-is:

```
Text # not a heading
```

I am happy for you to close this PR and land the character yourself with no attribution, or to reference it, whichever is less friction. The important thing is that #637 does not sit open waiting on a CLA for a one-character fix.

PR #639, opened 2026-09-08
Bump markdown-it to 15.0.1

joelzwarrington:
## Summary

Upgrades markdown-it from 12.3.2 to 15.0.1 and updates Markdoc's tokenizer/type layer to match its new API and shipped types.

- markdown-it 15 is bundled/ESM-first: no more `markdown-it/lib/*` subpath imports, so `Tokenizer` and the plugins (`annotations`, `comments`, `frontmatter`, `link`) now import types off the main `markdown-it` module (`MarkdownIt.StateBlock`, `MarkdownIt.MarkdownItOptions`, etc.).
- markdown-it now ships its own types, which removed the `markdown-it/lib/token` module path we were augmenting in `global.d.ts`. Replaced that global augmentation with a local `Token` type in `src/types.ts` that covers both real markdown-it tokens and Markdoc's own lightweight pseudo-tokens, and added the null-safety (`token.tag ?? ...`, `token.attrs ?? []`, etc.) the new optional fields require.
- Re-added the `allowIndentation` patch, regenerated for 15.0.1. markdown-it 15 ships bundled `dist/markdown-it.cjs.js`/`.mjs` instead of per-rule `lib/*.js` files, so the patch now guards the same 16 call sites in both bundled files instead of across nine small modules. Restored `patch-package` as a devDependency and the `prebuild` script that applies it.

## Why the patch is still needed

We looked at dropping the patch instead of porting it, since patches on bundled files are more brittle across upgrades. Without it, `marktest` fails: markdown-it treats 4+ spaces of leading indentation as a code block, which conflicts with content Markdoc needs to parse as regular blocks. Re-applying the patch unblocks this upgrade; longer-term we should look at upstreaming a proper option for this instead of patching bundled internals every version bump.

## Test plan
- [x] `npm run build` (patch applies cleanly via `prebuild`)
- [x] `npm test` — 267 specs, 0 failures
- [x] `npm run marktest` — passes with the patch applied


src/tokenizer/plugins/annotations.ts
```diff
@@ -98,10 +99,10 @@ function core(state: StateCore) {
   for (token of state.tokens) {
     if (token.type !== 'fence') continue;
 
-    if (token.info.includes(OPEN)) {
-      const start = token.info.indexOf(OPEN);
-      const end = findTagEnd(token.info, start);
-      const content = token.info.slice(start + OPEN.length, end);
+    if (token.info!.includes(OPEN)) {
```
maxblank-stripe: is `token.info` optional here? should we have a check above to prevent "Cannot read property includes of undefined"?

src/tokenizer/plugins/annotations.ts
```diff
@@ -126,11 +127,11 @@ function core(state: StateCore) {
     )
       continue;
 
-    token.children = parseTags(token.content, token.map[0]);
+    token.children = parseTags(token.content!, token.map![0]);
```
maxblank-stripe: same for `token.content` and `token.map`, how come they're optional now?

matv-stripe (approved):

yue-stripe (approved):

src/parser.ts
```diff
@@ -36,7 +34,7 @@ function annotate(node: Node, attributes: AttributeValue[]) {
 function handleAttrs(token: Token, type: string) {
   switch (type) {
     case 'heading':
-      return { level: Number(token.tag.replace('h', '')) };
+      return { level: Number((token.tag ?? '').replace('h', '')) };
```
maxblank-stripe: if `token.tag` is undefined, `level` would be `0`, which is invalid heading.

src/parser.ts
```diff
@@ -76,11 +74,12 @@ function handleAttrs(token: Token, type: string) {
 
         let align;
         if (attrs.style) {
-          if (attrs.style.includes('left')) {
+          const style = String(attrs.style);
```
maxblank-stripe, line 79: just curious, what's the TS type of `attrs.style`?

maxblank-stripe (changes requested):
Are the optional token properties coming from updated markdown-it library or because we defined them as optional in `src/types.ts`? My main concerns is using non-null assertion operator for those optional properties.

src/parser.ts
joelzwarrington: It's an `string | number`, `String()` normalizes to be only `string`.

CleanShot 2026-09-11 at 09 03 56
CleanShot 2026-09-11 at 09 04 46

src/parser.ts
joelzwarrington: Changed it to `token.tag!`.

src/tokenizer/plugins/annotations.ts
joelzwarrington: Yeah, this is in part an issue due to how we were previously typing this information through global module augmentation. I've updated this code so typing information is easier to understand.

src/tokenizer/plugins/annotations.ts
joelzwarrington: https://github.com/markdoc/markdoc/pull/639#discussion_r3990632804 helps explain this a bit, I believe this is also an issue caused by removal of global.d.ts module augmentation. So we have to do a bit more type checking on our end :)

Issue #321, opened 2023-02-01
Blockquotes with multiple children formatted incorrectly


emmatown:
### What happened?

When formatted, blockquotes only get a single `>` rather than a `>` on each line of content inside the blockquote.

### To reproduce

The following Markdoc file:
```markdown
> blockquote
> 
> with two paragraphs
```
formats to:
```markdown
> blockquote

with two paragraphs
```

### Version

0.2.1

mfix-stripe:
Thanks for catching this @emmatown! Will look into a fix immediately.

PR #640, opened 2026-09-11
Formatter: prefix every line of a blockquote's children with `> `

nsafai:
`format()` prefixes each child of a blockquote with `> ` on its first line only, so any child spanning several lines loses the quote on the rest. The multi-line sibling of #321, which fixed the two-paragraph case.

Input:

```
> Intro:
>
> - One
> - Two
```

Output on `main`:

```
> Intro:
> 
> - One
- Two
```

Nested blockquotes and fenced code inside a blockquote break the same way. Found via Keystatic, which writes files with `format()`: saving an essay with a quoted list silently restructured it.

Fix: format the children unindented, join them with a blank line, and prefix every line with `> ` at the blockquote's own indent, so a quote inside a list item still nests. Three tests added; all fail on `main`.

### DIFF
--- a/src/formatter.test.ts
+++ b/src/formatter.test.ts
@@ -833,6 +833,61 @@ ${'`'.repeat(4)}
     check(source, expected);
     stable(expected);
   });
+  it('lists in blockquotes', () => {
+    const source = `
+> Intro:
+>
+> - One
+> - Two
+>
+> 1. First
+> 2. Second
+>
+> After`;
+
+    const expected = `
+> Intro:
+> 
+> - One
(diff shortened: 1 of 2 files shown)


joelzwarrington (approved):
Thank you for the contribution!

nsafai:
> Thank you for the contribution! 

Happy to! Btw, thanks for the quick merge @joelzwarrington 

Any chance this could go out in a patch release soon? Hoping to bump Keystatic's Markdoc dependency to pick it up, since its editor writes files with format().

PR #642, opened 2026-09-11
Parse multi-line tag openings inside blockquotes

### LINKED ISSUE #641: Multi-line tag opening is not parsed inside a blockquote
A tag opening that spans several lines is not recognised inside a blockquote.

Outside a blockquote, this parses to a `callout` tag:

```
{% callout
   type="note" %}
Hi
{% /callout %}
```

Inside a blockquote, the same text parses to an `error` node, a paragraph, and a stray closing tag:

```
> {% callout
>    type="note" %}
> Hi
> {% /callout %}
```

```js
Markdoc.parse('> {% callout\n>    type="note" %}\n> Hi\n> {% /callout %}\n')
  .children[0].children.map((n) => n.type);
// ['error', 'paragraph', 'tag']
```

Tested on `main` (0.5.9).

This also affects `format()`: a tag with a long opening inside a blockquote is wrapped across lines (`maxTagOpeningWidth`), and the result no longer parses.

nsafai:
Fixes #641: a tag opening that spans several lines parses everywhere except inside a blockquote, where the `> ` markers of the continuation lines end up inside the tag text and the grammar rejects it.

This isn't just a hand-written edge case. `format()` wraps any tag opening longer than `maxTagOpeningWidth`, so a tool that formats on save (Keystatic, for one) turns a long callout inside a quote into an error node on the next parse.

Fix: read the tag's lines through `state.getLines`, which strips container markers the same way markdown-it's own block rules do. Single-line tags are unaffected. Test added beside the existing multi-line tag tests; it fails on `main`.

Independent of #640, which fixes the formatter side of the same round trip.

### DIFF
--- a/src/tokenizer/plugins/annotations.test.ts
+++ b/src/tokenizer/plugins/annotations.test.ts
@@ -167,6 +167,25 @@ describe('MarkdownIt Annotations plugin', function () {
         expect(example.length).toEqual(5);
         expect(example[2].children.length).toEqual(1);
       });
+
+      it('inside a blockquote', function () {
+        const example = parse(`
+        > {% test #foo .bar
+        >   baz=1 %}
+        > This is a test
+        > {% /test %}
+        `);
+
+        expect(example).toDeepEqualSubset([
+          { type: 'blockquote_open', nesting: 1 },
+          ...basicExample,
+          { type: 'blockquote_close', nesting: -1 },
+        ]);
+        expect(example[1].map).toDeepEqual([0, 2]);
+        expect(example[2].map).toDeepEqual([2, 3]);
+        expect(example[5].map).toDeepEqual([3, 4]);
+        expect(example.length).toEqual(7);
+      });
     });
 
     describe('inline', function () {
--- a/src/tokenizer/plugins/annotations.test.ts
+++ b/src/tokenizer/plugins/annotations.test.ts
@@ -186,6 +186,113 @@ describe('MarkdownIt Annotations plugin', function () {
         expect(example[5].map).toDeepEqual([3, 4]);
         expect(example.length).toEqual(7);
       });
+
+      it('followed by content after the blockquote', function () {
+        const example = parse(`
+        > {% test #foo .bar
+        >   baz=1 %}
+        > This is a test
+        > {% /test %}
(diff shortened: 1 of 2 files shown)


src/tokenizer/plugins/annotations.test.ts
joelzwarrington, line 1: Do you mind adding a few more test cases here? I'm thinking we might want to test the following scenarios to confirm:

1. List-item containers work with multi-line tags

```md
- {% test #foo .bar
        baz=1 %}
    This is a test
    {% /test %}
```

2. Nested containers with multi-line tags, can do either blockquotes → list item or vice-versa

```md
> - {% test #foo .bar
>     baz=1 %}
>   This is a test
>   {% /test %}
```

3. Parsing after a multi-line tag within a blockquote works as expected

```md
> {% test #foo .bar
>   baz=1 %}
> This is a test
> {% /test %}
After the blockquote
```

4. A single-line tag inside a blockquote works as expected

```md
> {% test #foo .bar baz=1 %}
> This is a test
> {% /test %}
```

joelzwarrington (changes requested):
Overall, changes look good to me, but would love to see a few more test cases that might catch some other issues. Let me know what you think!

src/tokenizer/plugins/annotations.test.ts
nsafai: @joelzwarrington  good ideas, just added these 4 tests, should be ready for ur review

joelzwarrington (approved):

src/tokenizer/plugins/annotations.test.ts
joelzwarrington: thank you for adding the tests @nsafai! ❤️

nsafai:
> Overall, changes look good to me, but would love to see a few more test cases that might catch some other issues. Let me know what you think!

Thanks for the quick review @joelzwarrington! Since this is approved now, any chance you could merge it, and then cut a patch release that includes both this and #640? 

I'm hoping to bump Keystatic's Markdoc dependency to pick up both blockquote fixes, since its editor writes files with format(), and the last published version (0.5.9) predates both. 

Happy to help if there's anything else needed to get a release out.

